PR #5, opened 2018-04-04
ATLAS-2530 trim query string before using it as name attribute for Hi…

bpgergo:
…ve process and column lineage


nixonrodrigues:
The fix already committed sometime back, so closing PR. Thanks @bpgergo for your PR

PR #12, opened 2018-10-17
ATLAS-2925: Better logging for graph database initialization error

gmcoringa:
Improve logs to know the reason of a graph failure. The old code was appending the exception object, which was only writing the exception message without the real reason of the graph database initialization failure.

This pr solves [ATLAS-2925](https://issues.apache.org/jira/browse/ATLAS-2925)


gmcoringa:
@mneethiraj done

mneethiraj:
+1

PR #18, opened 2018-10-26
Fixing logging of exceptions, other than NotificationException

maxneaga:
Currently, exceptions such as "Failed to construct kafka producer"
are not logged, because they are an instance of KafkaException.
This fix makes sure we log these exceptions as well.
image


maxneaga:
@apoorvnaik-hortonworks Is opening a pull request here the right way to get the changes incorporated into Atlas?

PR #19, opened 2018-11-01
Fix a bug in HBaseBasedAuditRepository.putEventsV2

TomDu:
Fix wrong function call.


TomDu:
@mneethiraj Hi, thanks for merging the PR and sorry for the late response. I don't have an account in JIRA, so I just created one, with the same user name. You can update the commit if needed :)

PR #21, opened 2018-12-14
ATLAS-3001: Fixed AtlasClientV2.updateClassifications() to not expect a return body

paul-k-civitas:
[JIRA](https://issues.apache.org/jira/browse/ATLAS-3001)

Right now, the `AtlasClientV2.updateClassifications()` method is not functional because it expects a response body that can be deserialized into an `AtlasClassifications` instance, but the [response according to the REST API](https://atlas.apache.org/api/v2/resource_EntityREST.html#resource_EntityREST_updateClassifications_PUT) will be empty, with a status code of `204 No Content`. This change fixes the client so that the response's body will not be de-serialized.


ashutoshm:
Changes look good.

nixonrodrigues:
This issue is already in ATLAS-3001,  Thanks @paul-k-civitas for reporting with PR.
closing this PR.

PR #22, opened 2019-01-30
Delete relationship fix

ccdigix:
ATLAS-3041: This pull request contains a fix to the problem whereby relationship type definitions cannot be deleted via the REST API.


nixonrodrigues:
This issue is fixed in https://issues.apache.org/jira/browse/ATLAS-3041.
Thanks @ccdigix for reporting the issue with PR. 

Closing PR.

PR #23, opened 2019-02-04
ability to enable audits in SSL enabled cassandra backend

chaitrarao04:
The existing code in CassandraBasedAuditRepository can write audits only to cassandra that does not have SSL enabled and credentials setup.

The fix in this PR reads the truststore certificate location, truststore password, cassandra username, cassandra password from the application properties and established connection to cassandra over SSL.

### DIFF
--- a/repository/src/main/java/org/apache/atlas/repository/audit/CassandraBasedAuditRepository.java
+++ b/repository/src/main/java/org/apache/atlas/repository/audit/CassandraBasedAuditRepository.java
@@ -20,9 +20,11 @@ package org.apache.atlas.repository.audit;
 import com.datastax.driver.core.BatchStatement;
 import com.datastax.driver.core.BoundStatement;
 import com.datastax.driver.core.Cluster;
+import com.datastax.driver.core.JdkSSLOptions;
 import com.datastax.driver.core.PreparedStatement;
 import com.datastax.driver.core.ResultSet;
 import com.datastax.driver.core.Row;
+import com.datastax.driver.core.SSLOptions;
 import com.datastax.driver.core.Session;
 import com.google.common.annotations.VisibleForTesting;
 import org.apache.atlas.AtlasException;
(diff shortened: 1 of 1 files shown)


repository/src/main/java/org/apache/atlas/repository/audit/CassandraBasedAuditRepository.java
```diff
@@ -217,7 +253,22 @@ void createSession() throws AtlasException {
     Cluster.Builder cassandraClusterBuilder = Cluster.builder();
 
     String hostname = APPLICATION_PROPERTIES.getString(CASSANDRA_HOSTNAME_PROPERTY, "localhost");
-    Cluster cluster = cassandraClusterBuilder.addContactPoint(hostname).withClusterName(clusterName).withPort(port).build();
+
+    Cluster cluster = null;
+    if (isSSLEnabled) {
+      if (StringUtils.isNotEmpty(username) && StringUtils.isNotEmpty(password)) {
```
gmcoringa, line 259: User credentials should not be locked into ssl, there are scenarios where only authentication is enought.

repository/src/main/java/org/apache/atlas/repository/audit/CassandraBasedAuditRepository.java
```diff
@@ -69,6 +83,17 @@
...
   public static final String CASSANDRA_REPLICATION_FACTOR_PROPERTY = "atlas.EntityAuditRepository.replicationFactor";
   public static final String CASSANDRA_AUDIT_KEYSPACE_PROPERTY = "atlas.EntityAuditRepository.keyspace";
+  
+  //flag to indicate if SSL is enabled for cassandra (server side SSL)
+  public static final String CASSANDRA_SSL_ENABLED = "atlas.graph.storage.sslEnabled";
+  public static final String CASSANDRA_SSL_TRUSTSTORE_LOCATION = "atlas.graph.storage.cassandra.ssl.truststore.location";
+  public static final String CASSANDRA_SSL_TRUSTSTORE_PASSWORD = "atlas.graph.storage.cassandra.ssl.truststore.password";
+
+  // cassandra login credentials
+  public static final String CASSANDRA_USERNAME = "atlas.graph.storage.username";
```
gmcoringa, line 93: The user credentials could use different properties, so you can have different users for audit and atlas keyspaces.

PR #25, opened 2019-02-25
EDATA-8538: Remove the predefined types that come with Atlas

smedberg:
⚠️ WIP/POC ⚠️ 

One possible approach to https://zendesk.atlassian.net/browse/EDATA-8538: Remove the predefined types that come with Atlas.

This PR removes the JSON files which specify the predefined types.  Numerous tests depend on those predefined tests, so those tests are also removed.  The removed tests have not been updated or replaced with tests that don't rely on the JSON data.


apoorv-naik:
Can you provide a justification around how Atlas would benefit from this? 

The Hadoop models are at the core of Atlas, plainly deleting these would cause a lot of deployments to break, one possible alternative can be to make use of a profile to build a "Hadoop" specific Atlas distro and only run the "dependent" tests when the profile is active. Also, instead of linking to the internal JIRA, it'd be better if you create an issue in GitHub or apache JIRA (preferred) with details about the changeset.

smedberg:
Oh my gosh, sorry, I intended to make this a PR on our private fork, not on the public repo.  In fact, my basic goal was to show that this is probably not a good approach!  I will close this momentarily- it was never intended to be public.

Sorry!

PR #30, opened 2019-03-22
ATLAS-3001 :- Fix AtlasClientV2.updateClassifications() throws serialization error due to wrong status code set in API

chaitalicod:
This patch provides fix to update **updateClassifications** method of **AtlasClientV2** with proper response status code which is returned by v2/entity/guid/%s/classifications API.

Added Testcase to validate the return response status code.


chaitalicod:
got committed to master, closing PR

PR #32, opened 2019-03-25
ATLAS-3092: Atlas Plugin ClassLoader Doesn't Restore Thread ClassLoader

rajkrrsingh:
**PROBLEM:** Atlas Plugin ClassLoader's MyComponent parent always set to UDFClassLoader after hiveserver2 start even though it is closed, Hiveserver2 can run sessions concurrently with different sessionstate, so if the sessionstate disgraded and associated UDF classloader is closed, MyComponent parent loader still has this closed parent classloader. this can yield into the CNF exception for custom hive udf.

**Fix:** In this proposed fix, during activate call I am storing current classloader in prevClassloader and during deactivate checking if the plugintype is hive( don't want to break the other components) then setting the Thread current class loader back to the previous one.

**Test:** Run few test on my reproduce where hive custom udf were failing due to CNF which are succeeded now.

### DIFF
--- a/plugin-classloader/src/main/java/org/apache/atlas/plugin/classloader/AtlasPluginClassLoader.java
+++ b/plugin-classloader/src/main/java/org/apache/atlas/plugin/classloader/AtlasPluginClassLoader.java
@@ -40,13 +40,15 @@ public final class AtlasPluginClassLoader extends URLClassLoader {
     private static volatile AtlasPluginClassLoader me = null;
 
     private final MyClassLoader componentClassLoader;
+    private final String pluginType;
+    private ClassLoader previousLoader;
 
     private AtlasPluginClassLoader(String pluginType, Class<?> pluginClass) throws URISyntaxException {
-        this(AtlasPluginClassLoaderUtil.getPluginImplLibPath(pluginType, pluginClass));
+        this(AtlasPluginClassLoaderUtil.getPluginImplLibPath(pluginType, pluginClass),pluginType);
     }
 
     //visible for testing
-    AtlasPluginClassLoader(String libraryPath) {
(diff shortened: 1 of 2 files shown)


plugin-classloader/src/main/java/org/apache/atlas/plugin/classloader/AtlasPluginClassLoader.java
```diff
@@ -206,7 +209,7 @@ public void activate() {
         if (LOG.isDebugEnabled()) {
             LOG.debug("==> AtlasPluginClassLoader.activate()");
         }
-
+        previousLoader = Thread.currentThread().getContextClassLoader();
```
mneethiraj, line 212: AtlasPluginClassLoader instance could be used in multiple threads simultaneously. So, saving 'previousLoader' in an instance member is not a safe option. @rmani (Ramesh Mani) is working on addressing the same issue in Ranger via RANGER-2376. I would suggest to wait for his Ranger fix and apply the same in Atlas as well.

shawnweeks: I wonder if it would be safe to save it on a per thread basis then?

shawnweeks: Or just handling this in the HiveHook it self? Then it should all be in the same thread and would be hive specific.

nixonrodrigues:
This issue is resolved in ATLAS-3092 and fix is merged, so closing the PR.
Thanks @rajkrrsingh for PR  and @shawnweeks @mneethiraj  for reviews.

PR #34, opened 2019-03-25
Fixed ability to skip CSRF filtering through configuration

calmera:
The previous implementation tried to read a string, not a string array. Apache Commons Configuration's behaviour when asking for a string in what essentially is an array defaults to returning the first element of the array.


apoorv-naik:
Looks good to me. Can you please open a JIRA for this ?

PR #36, opened 2019-03-25
ATLAS-3098: Missing getTypesDef with AtlasRelationshipDef param in AtlasTypeUtil.

ZepHakase22:
In AtlasTypesDef.java Atlas team introduced the AtlasTypesDef .ctor with the List<AtlasRelationshipDef> parameter but in AtlasTypeUtil the getTypesDef method does not accept 5 parameters.

I think we can use directly the AtlasTypesDef contructors so this Bug in my opinion is a minor issue. Any way I submitted the pull-request


ZepHakase22:
Resolved by ATLAS-3099.

PR #47, opened 2019-04-29
quick-start.py instead of quick-start.sh

wangmengqiang001:
it should be quick-start.py instead of quick-start.sh


wangmengqiang001:
Another opened issue is merged by mistake

wangmengqiang001:
OK, it is no proble, i am sure now!

PR #49, opened 2019-04-29
ATLAS-3180:New Bulk retrieval API with servicetype parameter

ZepHakase22:
ATLAS-3180: The Bulk api retrieval returns all defined types or filtered types by name, type and supertype. With the addition of the service type among the basic attributes of the Atlas types it is necessary to have the possibility to search also for service type through a http request of the type:

```http://host:port/api/atlas/V2/types/typedefs?serviceType = requested-service-type```

to search only for types related to a specific service type.
This pull request solves this problem


ZepHakase22:
@mneethiraj - Thx. Please it is important that you supervise the ATLAS-3227.  It's a bug in the bulk delete.  I found a solution and made the pull request without any comment from the team. Maybe you don't like the solution but for me it is important to fix the bug however also for 1.2.0.

PR #58, opened 2019-06-07
[ATLAS-3261] Set kafka user as current principal for Ranger Authorization

arempter:
If lineage message sent to ATLAS_HOOK has non empty "user" value, NotificationHookConsumer.java will set Authentication context to this user, allowing check in Ranger policies by Atlas authorizer.

Currently no user is set by NotificationHookConsumer, so Atlas authorizer is not invoked.

### DIFF
--- a/webapp/src/main/java/org/apache/atlas/notification/NotificationHookConsumer.java
+++ b/webapp/src/main/java/org/apache/atlas/notification/NotificationHookConsumer.java
@@ -486,7 +486,7 @@ public class NotificationHookConsumer implements Service, ActiveStateChangeHandl
 
         private void setMessagePrincipal(String messageUser) throws AtlasException {
             Configuration configuration = ApplicationProperties.get();
-            boolean msgPrincipalEnabled = configuration.getBoolean("atlas.authentication.method.kafka.message.principal", false);
+            boolean msgPrincipalEnabled = configuration.getBoolean("atlas.notification.authorize.using.message.user", false);
 
             if(msgPrincipalEnabled) {
                 if (!StringUtils.isEmpty(messageUser)) {
--- a/webapp/src/main/java/org/apache/atlas/notification/NotificationHookConsumer.java
+++ b/webapp/src/main/java/org/apache/atlas/notification/NotificationHookConsumer.java
@@ -77,6 +77,12 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.context.annotation.DependsOn;
 import org.springframework.core.annotation.Order;
+import org.springframework.security.authentication.UsernamePasswordAuthenticationToken;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.GrantedAuthority;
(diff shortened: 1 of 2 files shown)


webapp/src/main/java/org/apache/atlas/notification/NotificationHookConsumer.java
```diff
@@ -477,6 +484,31 @@ public void doWork() {
...
+                    UserDetails principal;
+
+                    if (grantedAuths.isEmpty()) {
+                        principal = new User(messageUser, "", new HashSet<>());
+                    } else {
+                        principal = new User(messageUser, "", grantedAuths);
+                    }
+
+                    Authentication finalAuthentication = new UsernamePasswordAuthenticationToken(principal, "");
+                    SecurityContextHolder.getContext().setAuthentication(finalAuthentication);
```
nixonrodrigues, line 503: This will be adding authentication in context for every message and this can cause performance issue. Can we maintain session some how ?

arempter: I think it is difficult to say about the session here, since every kafka message processed can potentially have different user or properties. 
I see it more similar in nature to REST call, where you authenticate every request.

arempter:
> @arempter - the identity of the Kafka message producer is not available at the consumer side; Kafka doesn't support this. Though Atlas notification has a filed named 'user', this can't be used for authorization, as it trusts the senders to set correct value. Hence it is very important that ACLs on ATLAS_HOOKS Kafka topic is set carefully - to allow only trusted users to produce messages.
> 
> In short, notification mechanism must only be used for trusted users. For usecases that involve calls from untrusted users, which require authorizations, REST APIs should be used.

I completely agree, that there is no good way to guarantee that user name in message is correct and validated.
I see it more like addition to topic ACL, which is main mechanism that grants access for posting messages.
The issue with ACL is that is all-or-noting approach, without fine grained control of what user can do in atlas. 
On the other hand even if user will set any value for the user field, there is still external service (ranger policy) which is being checked and user cannot setup policy himself.

mneethiraj:
@arempter - I agree that authorizing based on username in the message can provide an additional layer of control. Given that authentication for each message is likely to add a sizable overhead, I would suggest to make this additional, per-message, authorization an option - controlled via a configuration like "atlas.notification.authorize.using.message.user". This will help avoid the overhead for deployments that don't need this additional layer of authorization control.

arempter:
> like "atlas.notification.authorize.using.message.user". This will help avoid the overhead for deployments that don't need this additional layer of authorization control.

Sure, updated parameter name

arempter:
@mneethiraj is there something more that should be done here?

bolkedebruin:
Ping @mneethiraj @nixonrodrigues

mneethiraj:
@bolkedebruin, @arempter - the patch looks good. I rebased this patch with latest master, and added caching of authentication objects; please review the updated patch at https://reviews.apache.org/r/71841/.

PR #61, opened 2019-06-21
ATLAS-3295: Remove unused setup code and documentation

andrewluotechnologies:
It seems this setup code was added several years back to resolve some concurrency issues, and partly removed later on.  However, the documentation does not reflect this, and the startup script still refers to a now-deleted class.

### DIFF
--- a/dashboardv2/public/css/scss/loader.scss
+++ b/dashboardv2/public/css/scss/loader.scss
@@ -156,7 +156,8 @@
     background: linear-gradient(to right, #eeeeee 8%, #dddddd 18%, #eeeeee 33%);
     background-size: 700px 10px;
     border-radius: 8px;
-
+    outline: 1px #fff solid;
+    outline-offset: -1px;
 }
 
 @keyframes placeHolderLoader {
--- a/dashboardv2/public/css/scss/table.scss
+++ b/dashboardv2/public/css/scss/table.scss
@@ -244,8 +244,6 @@ td {
             height: 20px;
             margin: 0px 5px;
             float: left;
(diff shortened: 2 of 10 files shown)


distro/src/bin/atlas_start.py
```diff
@@ -143,13 +135,10 @@ def main():
     web_app_path = os.path.join(web_app_dir, "atlas")
     if (mc.isCygwin()):
         web_app_path = mc.convertCygwinPath(web_app_path)
-    if not is_setup:
-        start_atlas_server(atlas_classpath, atlas_pid_file, jvm_logdir, jvm_opts_list, web_app_path)
-        mc.wait_for_startup(confdir, 300)
-        print "Apache Atlas Server started!!!\n"
-    else:
-        process = mc.java("org.apache.atlas.web.setup.AtlasSetup", [], atlas_classpath, jvm_opts_list, jvm_logdir)
```
thealmightygrant: To go along with this, the current script fails. And this line needs to be changed to:
```py
process = mc.java("org.apache.atlas.web.setup.AtlasSetupModule", [], atlas_classpath, jvm_opts_list, jvm_logdir)
```
otherwise, `-setup` fails.

andrewluotechnologies: I don't believe there is a AtlasSetupModule anymore: https://github.com/apache/atlas/tree/master/webapp/src/main/java/org/apache/atlas/web/setup

The purpose of this PR is to entirely remove the -setup option as it doesn't seem to be needed anymore.

PR #62, opened 2019-06-21
ATLAS-3296: Correct elasticsearch hostname property in atlas-application.properties

andrewluotechnologies:
The property name that is commented out in the default atlas-application.properties is incorrect.


sarathsubramanian:
+1 thanks for the patch @andrewluotechnologies

andrewluotechnologies:
This has been committed in a876e697c3c0c936e7efc33cecbe1e1c32a6be20

PR #66, opened 2019-08-16
fix - aws_s3_bucket createTime attribute

kr7ysztof:
It fixes the inconsistency with name attribute `createtime` - all attribute names are camelCase


bolkedebruin:
Cc @sarathkumarsubramanian

bolkedebruin:
https://issues.apache.org/jira/browse/ATLAS-3393

PR #75, opened 2019-11-08
ATLAS-3519: pool performance in hive lineage query

zhangxiong0301:
when query lineage info for hive, the throughput reachs only 3 query per second.

we use jstack and jfr+flame graph tools and found that the  GremlinGroovyScriptEngine object loading class  logic takes much of the time. so we make the GremlinGroovyScriptEngine a global singleton object, which would be shared accross all query threads. finally the query performance improved much and got 200 qps.


bolkedebruin:
@zhangxiong0301 Usually the Atlas folks use the Apache review board to review patches. Although we should maybe request them to start using github PRs as well cc @mneethiraj

graphdb/janus/src/main/java/org/apache/atlas/repository/graphdb/janus/AtlasJanusGraph.java
```diff
@@ -330,13 +332,16 @@ public void exportToGson(OutputStream os) throws IOException {
 
     @Override
     public GremlinGroovyScriptEngine getGremlinScriptEngine() {
+        return scriptEngine;
```
mneethiraj: GremlinGroovyScriptEngine returned here may not safe for use in multiple threads simultaneously i.e. side-effects of an execution in a thread might be visible in other threads. I suggest to use a ThreadLocal<GremlinGroovyScriptEngine>, to instantiate and use one script engine per thread.

leslizhang: according to tinkerpop github code,the GremlinGroovyScriptEngine class is thread-safe.below is the link of tinkerpop test code for thread-safe feature ( shouldBeThreadSafe() method ):
https://github.com/apache/tinkerpop/blob/master/gremlin-groovy/src/test/java/org/apache/tinkerpop/gremlin/groovy/jsr223/GremlinGroovyScriptEngineOverGraphTest.java
thanks.

mneethiraj:
@zhangxiong0301  - though the script engine class itself is thread-safe, it can contain state (like variable set at the engine scope) and having this thread updated/read in multiple threads simultaneously can result in unexpected results. Hence the suggestion to use ThreadLocal<>, to ensure that an engine is not used simultaneously by multiple threads.

zhangxiong0301:
> @zhangxiong0301 - though the script engine class itself is thread-safe, it can contain state (like variable set at the engine scope) and having this thread updated/read in multiple threads simultaneously can result in unexpected results. Hence the suggestion to use ThreadLocal<>, to ensure that an engine is not used simultaneously by multiple threads.
your proposal is reasonable,and i have made the scriptengine threaddd local.

mneethiraj:
@zhangxiong0301  - the patch looks good! This will be merged in master branch shortly. Thanks!

PR #79, opened 2019-12-18
ATLAS-3567 - Bypass package-lock.json error during build

hayssams:
Disabling package-lock.json Locally to avoid failure during build  "npm notice created a lockfile as package-lock.json. You should commit this file."


nixonrodrigues:
+1 for the patch. @hayssams thanks for fix.

PR #83, opened 2020-01-24
Fixing dev-support/atlas-docker/Dockerfile image wouldn't build / run

v-odier-heetch:
The dockerfile in ./dev-support/atlas-docker/ doesn't build because the install is unable to check all licenses.

Fixed by adding -Drat.skip=true to `MAVEN_OPTS`

Once built the image won't start because solr crash instantly. This is because solr doesn't like to run as root and need -force flag to do so. To avoid edditing the start script a user "ubuntu" has been added in the Dockerfile.

Also added 2 build options describe in ./dev-support/atlas-docker/readme.md

Option `--build-arg MVN_JOB=8` to setup -T option when running mvn. Default is 1  
Option `--build-arg BRANCH=master` to setup branch to clone. Default is master

`MVN_JOB` was added because the build takes ages  
`BRANCH` because it makes it easier to build other branch (i needed to build branch-2.0)

### DIFF
--- a/dev-support/atlas-docker/Dockerfile
+++ b/dev-support/atlas-docker/Dockerfile
@@ -16,6 +16,13 @@
 
 FROM ubuntu:18.04
 
+# --build-arg MVN_JOB=8 to built with option -T 8 default to 1
+ARG MVN_JOB=1
+# --build-arg BRANCH=master to build specific branc default to master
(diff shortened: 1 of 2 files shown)


v-odier-heetch:
i've realized some path are hardcoded closing until fixed

PR #90, opened 2020-03-05
[ATLAS-3654] enable support for Solr in standalone/http mode

dwarszawski:


mneethiraj:
This patch was merged, and changes are available since Apache Atlas 2.1.0.

PR #91, opened 2020-03-05
ATLAS-3655: Create 'spark_application' type to avoid 'spark_process' from being updated for multiple operations

vladhlinsky:
## What changes were proposed in this pull request?

Create `spark_application` type to avoid `spark_process` from being updated for multiple operations. Currently, Spark Atlas Connector uses `spark_process` as a top-level type for a Spark session, thus it's being updated for multiple operations within the same session.

The following statements:
```
spark.sql("create table table_1(col1 int,col2 string)");
spark.sql("create table table_2 as select * from table_1");
```
result in the next correct lineage:
```
table1 ------> spark_process1 -------> table2
```
but executing similar statements in the same spark session:
```
spark.sql("create table table_3(col1 int,col2 string)"); 
spark.sql("create table table_4 as select * from table_3");
```
result in the same `spark_process` being updated and the lineage now connects all the 4 tables.
The proposal is to create a `spark_application` entity and associate all `spark_process` entities (created within that session) to it.

## How was this patch tested?

Manually using modified version of Spark Atlas Connector:
- Installed and started Atlas.
- Executed the next statements using spark-shell:

```
spark.sql("create table table_1_17(col1 int,col2 string)");
spark.sql("create table table_2_17 as select * from table_1_17");
spark.sql("create table table_3_17(col1 int,col2 string)");
spark.sql("create table table_4_17 as select * from table_3_17");
```

- Verified that all 4 entites are connected in Atlas lineage.
- `1100-spark_model.json` is updated with proposed changes.
- Once again executed similar statements:

```
spark.sql("create table table_1_37(col1 int,col2 string)");
spark.sql("create table table_2_37 as select * from table_1_37");
spark.sql("create table table_3_37(col1 int,col2 string)");
spark.sql("create table table_4_37 as select * from table_3_37");
```

- Verified that two `spark_process` entities are created,
that have a single `spark_application` entity as `application`.
Each of these processes has it's own lineage.

PR #94, opened 2020-03-13
ATLAS-3665 Add 'queryText' attribute to the 'spark_process' type

vladhlinsky:

Add `queryText` attribute to the `spark_process` type in order to make `spark_process` more readable by the user. The `queryText` attribute stores exact SQL query that is executed within Spark session as a `spark_process`.

## How was this patch tested?
    
Manually using modified version of Spark Atlas Connector:
- Install and start Atlas.
- Stop Atlas.
- Copy `015-spark_process_add_query_text_attribute.json` to `models/1000-Hadoop/patches`.
- Start Atlas.
- Verify that patch is applied and `application.log` contains the following entry:
```
INFO  - [main:] ~ TYPEDEF_PATCH_1000_015_001 (status: APPLIED; action: ADD_ATTRIBUTE) in file: ...
```
- Execute the next statements using spark-shell:
```
spark.sql("create table source_2_77_10(id int, new_id int, name string)");
spark.sql("create table ctas_2_77_10 as select id as new_id from source_2_77_10");
```
- Verify that corresponding `spark_process` is created and has `queryText`:
```
create table ctas_2_77_10 as select id as new_id from source_2_77_10
```

### DIFF
--- a/addons/models/1000-Hadoop/patches/015-spark_process_add_query_text_attribute.json
+++ b/addons/models/1000-Hadoop/patches/015-spark_process_add_query_text_attribute.json
@@ -1,0 +1,23 @@
+{
+  "patches": [
+    {
+      "id": "TYPEDEF_PATCH_1000_015_001",
+      "description": "Add 'queryText' attribute to spark_process",
+      "action": "ADD_ATTRIBUTE",
+      "typeName": "spark_process",
+      "applyToVersion": "1.0",
+      "updateToVersion": "1.1",
+      "params": null,
+      "attributeDefs": [
+        {
+          "name": "queryText",
+          "typeName": "string",
+          "cardinality": "SINGLE",
+          "isIndexable": true,
+          "isOptional": true,
+          "isUnique": false
+        }
+      ]
+    }
+  ]
+}


vladhlinsky:
Attaching screenshots.
Screenshot from 2020-03-13 15-28-00
Screenshot from 2020-03-13 15-28-13

vladhlinsky:
cc @HeartSaVioR @sarathsubramanian

HeartSaVioR:
How it will be changed after adopting #91? Would it be possible to couple query with each spark_process after #91, or we won't know about relationship between spark_process and the query? 

If it's former, ideally `spark_process` would have single query instead of queries. If it's latter, recentQueries should be added into `spark_application` instead of `spark_process`.

vladhlinsky:
@HeartSaVioR thank you for the review!
After adopting #91 Spark Atlas Connector will create a `spark_process` for each supported at this time `QueryExecution` event except of `CreateTableCommand` and `CreateDataSourceTableCommand`. Most of these commands could be mapped to single query. Since `CreateTableCommand`, `CreateDataSourceTableCommand` and `ExternalCatalogEvent`'s do not result in a `spark_process` being created, it's possible to keep `recentQueries` attribute as list to aggregate corresponding SQL queries until a spark process is not created:
Screenshot from 2020-03-16 19-51-09

But it seems that at this moment queries that are mapped to `QueryExecution` events will be more suitable to make `spark_process` more readable by the user.

Previously, I referred to `recentQueries` attribute of the `hive_process` type, but as it turned out there is no case when it's value is a list of multiple queries:
https://github.com/apache/atlas/blob/master/addons/hive-bridge/src/main/java/org/apache/atlas/hive/bridge/HiveMetaStoreBridge.java#L377
https://github.com/apache/atlas/blob/master/addons/hive-bridge/src/main/java/org/apache/atlas/hive/hook/events/BaseHiveEvent.java#L663

The PR is updated to add single `queryText` attribute to the `spark_process` type.

vladhlinsky:
I checked behavior of the Spark Atlas Connector with proposed changes after adopting #91. Below is the list of cases when SAC creates a `spark_process` along with sample values of the `queryText` attribute:
1. ML process, streaming query process

**`queryText` attribute is not set.**

2. `InsertIntoHiveTable`, `InsertIntoHadoopFsRelationCommand`

**Tested with:**
```
spark.sql("create table table_3(id int,name string)");
spark.sql("insert into table_3 values(1,'John')");
```
**`queryText`:**
Screenshot from 2020-03-17 18-32-36

3. `CreateHiveTableAsSelectCommand`, `CreateDataSourceTableAsSelectCommand`

**Tested with:**
```
spark.sql("create table source_table_1(col1 int,col2 string)");
spark.sql("create table dest_table_1 as select * from source_table_1");
```
**`queryText`:**
Screenshot from 2020-03-17 18-36-29

4. `LoadDataCommand`

**Tested with:**
```
import java.io.{FileOutputStream, PrintWriter}
import java.nio.file.Files

val file = Files.createTempFile("input", ".txt").toFile
val out = new PrintWriter(new FileOutputStream(file))
out.write("a\nb\nc\nd\n")
out.close()

spark.sql(s"CREATE TABLE hive_table_7 (name string) USING HIVE")
spark.sql(s"LOAD DATA LOCAL INPATH '${file.getAbsolutePath}' OVERWRITE INTO TABLE hive_table_7")
```
**`queryText`:**
Screenshot from 2020-03-17 18-38-55

5. `InsertIntoHiveDirCommand`

**Tested with:**
```
spark.sql("create table overwrite_source_table_1(id int,name string)");
spark.sql(s"insert overwrite directory 'tempDir1' select * from overwrite_source_table_1")
```
**`queryText`:**
Screenshot from 2020-03-17 18-39-51

6. `CreateViewCommand`

**Tested with:**
```
spark.sql("create table view_source_table(col1 int,col2 string)");
spark.sql("create view dest_view as select * from view_source_table");
```
**`queryText`:**
Screenshot from 2020-03-17 18-40-42

7. `SaveIntoDataSourceCommand`, `WriteToDataSourceV2Exec`

**Tested with:**
```
val lines = Seq("SaveIntoDataSourceCommand").toDF("line")
val url = "jdbc:derby:memory:;databaseName=/tmp/test;create=true"
val requiredOpts = Map("url" -> url, "dbtable" -> "lines")
import org.apache.spark.sql.SaveMode.Overwrite
lines.write.options(requiredOpts).format("jdbc").mode(Overwrite).save
```
**`queryText` attribute is not set.**

`CreateTableCommand`, `CreateDataSourceTableCommand` do not result in a `spark_process` being created.

sarathsubramanian:
@vladhlinsky , can we use the patch files to add the new attribute to 'spark_process'? Existing deployments with spark_model needs to be updated to add this new attribute. You can take a look at [](https://github.com/apache/atlas/blob/master/addons/models/1000-Hadoop/patches/006-kafka_topic_add_attribute.json) for reference.

vladhlinsky:
@sarathsubramanian  thank you for the review! The PR is updated to use patch file.

sarathsubramanian:
Thanks for the patch @vladhlinsky. Changes looks good. +1

nixonrodrigues:
Thanks @vladhlinsky for PR, 
Thanks @HeartSaVioR @sarathsubramanian for review.

PR #96, opened 2020-04-26
[ATLAS-3756] Error when compile from source code - missing repo for restlet

hsaputra:
Error when compile from source code - can not find restlet for transitive dep

Need to add repositories where restlet available.


mneethiraj:
@hsaputra  - thank you for the patch. This fix is merged in master and branch-2.0 branches.

hsaputra:
Thanks @mneethiraj

PR #98, opened 2020-04-27
[ATLAS-3800] AWS scheme is missing the aws account id that contains the S3 bucket.

AmeerAssi:
adding optional field for account id, so it will not break existing products that are using the current scheme, and will add ability for products to have the context of which account contains the bucket.

### DIFF
--- a/addons/models/3000-Cloud/3030-aws_s3_v2_typedefs.json
+++ b/addons/models/3000-Cloud/3030-aws_s3_v2_typedefs.json
@@ -259,7 +259,7 @@
           "isIndexable": true,
           "isOptional":  true,
           "isUnique":    false,
-          "searchWeight" : 5
+          "searchWeight" : 10
         }
       ]
     }
--- a/addons/models/3000-Cloud/3030-aws_s3_v2_typedefs.json
+++ b/addons/models/3000-Cloud/3030-aws_s3_v2_typedefs.json
@@ -195,7 +195,7 @@
       "serviceType": "aws",
       "typeVersion": "1.0",
       "attributeDefs": [
-       {
+        {
           "name":        "accountId",
           "typeName":    "string",
           "cardinality": "SINGLE",
@@ -204,7 +204,7 @@
           "isUnique":    false,
           "searchWeight" : 5
         },
-		{
+        {
           "name":        "creationTime",
           "typeName":    "date",
           "cardinality": "SINGLE",
--- a/addons/models/3000-Cloud/3030-aws_s3_v2_typedefs.json
+++ b/addons/models/3000-Cloud/3030-aws_s3_v2_typedefs.json
@@ -196,15 +196,6 @@
       "typeVersion": "1.0",
       "attributeDefs": [
         {
-          "name":        "accountId",
-          "typeName":    "string",
-          "cardinality": "SINGLE",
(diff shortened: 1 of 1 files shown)


sidharthkmishra:
As per my understanding accountId is used to construct ARN and it may be used at AWS to distinguish resources. Could you please explain how will this be useful as Atlas metadata and what  is the use-case.

AmeerAssi:
So we are using Atlas models in our product.
the product gives scan capabilities for customer's S3 in AWS.
when scan result shown based on the model, it is missing for which account this bucket belongs. so it will be easy to access the account and review the scanned information by the customer.

AmeerAssi:
Thanks @sidharthkmishra for approving.
@sarathsubramanian @ashutoshm can you please take a look per Sidharth's.
I am not getting any response from you in mail.

sarathsubramanian:
@AmeerAssi for existing deployments to add this attribute, can you please include a patch file to add this attribute? You can check a similar patch for reference - https://github.com/apache/atlas/blob/master/addons/models/3000-Cloud/patches/004-s3_add_attributes.json

addons/models/3000-Cloud/3030-aws_s3_v2_typedefs.json
```diff
@@ -251,6 +251,15 @@
...
           "isUnique":    false
+        },
+        {
+          "name":        "accountId",
+          "typeName":    "string",
+          "cardinality": "SINGLE",
+          "isIndexable": true,
+          "isOptional":  true,
+          "isUnique":    false,
+          "searchWeight" : 5
```
sarathsubramanian: if you want this attribute to be searchable in quicksearch/suggestions. Add a higher value for search weight. 9 or 10

AmeerAssi:
> @AmeerAssi for existing deployments to add this attribute, can you please include a patch file to add this attribute? You can check a similar patch for reference - https://github.com/apache/atlas/blob/master/addons/models/3000-Cloud/patches/004-s3_add_attributes.json

@sarathkumarsubramanian 
AFAIK this model was never released with any Atlas version, can you please explain why this patch file is needed, as my understanding this patch file came for backward compatibility.

addons/models/3000-Cloud/3030-aws_s3_v2_typedefs.json
AmeerAssi: @sarathkumarsubramanian thanks for the suggestion.
done.

AmeerAssi:
@sarathkumarsubramanian kindly reminder for this review.

sarathsubramanian:
+1

sarathsubramanian:
@AmeerAssi can you create an apache jira and attach your patch in the JIRA. I will merge these changes. I need a apache jira id for commit message.

AmeerAssi:
Thank you.
Opened this issue:
https://issues.apache.org/jira/browse/ATLAS-3800

AmeerAssi:
Thank you.
Opened this issue:
https://issues.apache.org/jira/browse/ATLAS-3800

@sarathkumarsubramanian let me know if anything else is needed.

sarathsubramanian:
merged the changes. Please resolve this PR.

AmeerAssi:
Thank you.
Closing the PR

PR #111, opened 2020-09-16
[ATLAS-3906]: fix dependency conflict issue

Bing-ok:
[ATLAS-3906](https://issues.apache.org/jira/browse/ATLAS-3906) fix dependency conflict issue for org.apache.curator:curator-client:jar

### DIFF
--- a/server-api/pom.xml
+++ b/server-api/pom.xml
@@ -39,6 +39,7 @@
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-common</artifactId>
+            <version>3.3.0</version>
             <exclusions>
                 <exclusion>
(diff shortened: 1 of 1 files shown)


server-api/pom.xml
```diff
@@ -39,6 +39,7 @@
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-common</artifactId>
+            <version>3.3.0</version>
```
martin-g, line 42: The version is defined in the parent pom.
And it is updated to 3.3.0 with 80b305f827092f5e006d556aba3e02acee052c2d

martin-g: This PR could be closed

Bing-ok:
Thanks for your support! @martin-g

PR #123, opened 2021-03-04
Upgrade frontend-maven-plugin to support s390x

chenqiny:
frontend-maven-plugin 1.4 does not support s390x, takes s390x arch as x86, and failed to install node.js.

After 1.5, frontend-maven-plugin starts to support s390x.

With this change, mvn can build atlas on s390x and atlas can be executed on s390x environment.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1788,7 +1788,7 @@
                 <plugin>
                     <groupId>com.github.eirslett</groupId>
                     <artifactId>frontend-maven-plugin</artifactId>
-                    <version>1.4</version>
+                    <version>1.5</version>
                 </plugin>
 
                 <plugin>


pom.xml
```diff
@@ -1788,7 +1788,7 @@
                 <plugin>
                     <groupId>com.github.eirslett</groupId>
                     <artifactId>frontend-maven-plugin</artifactId>
-                    <version>1.4</version>
+                    <version>1.5</version>
```
martin-g: Even better upgrade it to the latest version - 1.11.2

chenqiny: Yes. I updated it.

chenqiny:
I tested the latest com.github.eirslett to latest 1.11.2 on both x86 and s390.

There is no compilation error. Is it possible review again? Thanks.

martin-g (approved):
LGTM

PR #130, opened 2021-04-10
ATLAS-4318 : fix python client api url concatenation

claudio-benfatto:
Some of the python client api endpoints are currently broken as the result of the concatenation of the base url with the relative context path items.

Specifically this happens for those path which contain a trailing forward slash (`/`), e.g.
- https://github.com/apache/atlas/blob/master/intg/src/main/python/apache_atlas/client/entity.py#L61 and
- https://github.com/apache/atlas/blob/master/intg/src/main/python/apache_atlas/client/entity.py#L58

This produced requests of the kind:
```
/api/atlas/v2/entity/uniqueAttribute/type//topic_schema
```
which obviously resulted in errors.

Instead of simply removing the `/` from the base paths I've chosen the approach of replacing the url concatenation method with a more robust one that handles leading and trailing slashes transparently.

I did not find any tests for this module. If I'm wrong please let me know and I'll add/modify them.

Regards!

### DIFF
--- a/intg/src/main/python/apache_atlas/utils.py
+++ b/intg/src/main/python/apache_atlas/utils.py
@@ -17,6 +17,7 @@
 # limitations under the License.
 import enum
 import time
+from functools import reduce
 
 BASE_URI = "api/atlas/"
 APPLICATION_JSON = 'application/json'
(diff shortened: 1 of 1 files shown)


nikhilbonte (approved):
Looks goo to me

nixonrodrigues:
+1 for the PR

PR #139, opened 2021-06-08
ATLAS-4323: Add keys to Kafka notifications

bellemare:
In draft to seek feedback:

Important points:
1) Key is fixed to String, which is the format of GUID.
2) Keys will be GUID based, as the GUID field does not change. This will ensure data locality within the Kafka topic's partitions.
3) Currently, no support for arbitrary keys. It is not clear to me that there are any use-cases that would require us to provide the user with the ability to key against arbitrary fields / data structures.


mneethiraj:
- adding a field for 'key' in HookNotification and EntityNotification might be a simpler/smaller change. I suggest to consider this  approach.
- how are the keys generated, and where are they assigned to each notification being sent? I could see this in unit tests. Can you please point to appropriate code lines?

webapp/src/main/java/org/apache/atlas/notification/EntityNotificationListenerV2.java
```diff
@@ -155,14 +155,16 @@ public void onLabelsAdded(AtlasEntity entity, Set<String> labels) throws AtlasBa
...
         MetricRecorder metric = RequestContext.get().startMetricRecord("entityNotification");
-        List<EntityNotificationV2> messages = new ArrayList<>();
+        List<KeyValue<String, EntityNotificationV2>> messages = new ArrayList<>();
 
         for (AtlasEntity entity : entities) {
             if (isInternalType(entity.getTypeName())) {
                 continue;
             }
+            messages.add(new KeyValue<>(
+                             entity.getGuid(),
```
bellemare, line 165: @mneethiraj Adding GUID as key for entity here.

webapp/src/main/java/org/apache/atlas/notification/EntityNotificationListenerV2.java
```diff
@@ -171,21 +173,23 @@ private void notifyEntityEvents(List<AtlasEntity> entities, OperationType operat
...
+        List<KeyValue<String, EntityNotificationV2>> messages = new ArrayList<>();
 
         for (AtlasRelationship relationship : relationships) {
             if (isInternalType(relationship.getTypeName())) {
                 continue;
             }
 
-            messages.add(new EntityNotificationV2(toNotificationHeader(relationship), operationType, RequestContext.get().getRequestTime()));
+            messages.add(new KeyValue<>(
+                    relationship.getGuid(),
```
bellemare, line 184: @mneethiraj Adding GUID as key for relationship here.

notification/src/main/java/org/apache/atlas/notification/spool/Spooler.java
```diff
@@ -100,23 +102,27 @@ boolean write(List<String> messages) {
...
         boolean ret = false;
 
         try {
             byte[]     lineSeparatorBytes = SpoolUtils.getLineSeparator().getBytes(SpoolUtils.DEFAULT_CHAR_SET);
             DataOutput pw                 = indexManagement.getSpoolWriter();
 
-            for (String message : messages) {
-                pw.write(message.getBytes(SpoolUtils.DEFAULT_CHAR_SET));
+            for (KeyValue<String,String> message : messages) {
+                //TODO - Bellemare - Serialize this with a proper format
```
bellemare, line 122: @mneethiraj Ah yes, I think this would be an area where simply placing the key in HookNotification may make life a lot easier.

bellemare:
>     * adding a field for 'key' in HookNotification and EntityNotification might be a simpler/smaller change. I suggest to consider this  approach.

I'm trying this now, but it doesn't seem like it works conceptually.

For instance, it seems that `HookNotification` is extended by `EntityCreateRequestV2`. This extension, however, seems to pack in multiple entities via `AtlasEntitiesWithExtInfo`, meaning that there is no single key that `HookNotification` can use.

I think I need some help conceptually here: Is `HookNotification` and `EntityNotification` supposed to be a bundle of an arbitrary number of entities? I was originally under the impression that there were meant to be 1:1 (eg: one EntityNotification per entity).

PR #140, opened 2021-06-21
fix rename table with diff db bugs

CanGuan:
Fix rename table bugs like: ALTER TABLE db_name.test_1 RENAME TO db_name1.test_1;


rkundam (commented):
Can you please provide more context about the usecase and how this patch helps?

CanGuan:
> Can you please provide more context about the usecase and how this patch helps?

When user rename hive table with other db name by executing:
`ALTER TABLE db1.table RENAME TO db2.table`

HMS hook will send `ALTERTABLE_PROPERTIES` event msg instead of `ALTERTABLE_RENAME` (HiveMetastoreHookImpl#onAlterTable), causing old table still cant be found.

mehtaanshul:
Issue link - https://issues.apache.org/jira/browse/ATLAS-4358

PR #142, opened 2021-07-13
ATLAS-4358 move addition of default java patches to applyAll method

mehtaanshul:


mehtaanshul:
Issue link - https://issues.apache.org/jira/browse/ATLAS-4358

mehtaanshul:
@nixonrodrigues

nixon-atlan:
cc : @ashutoshm , can you review this ?

ashutoshm:
@mehtaanshul Thanks for the PR. I am in the process of verifying this change.

sarathsubramanian:
@mehtaanshul , can you add an explanation on why we need to move the java patch registration from init() to applyAll() method?

sarathsubramanian:
missed the details from JIRA, ignore my prev comment.

repository/src/main/java/org/apache/atlas/repository/patches/AtlasPatchManager.java
```diff
@@ -48,28 +48,24 @@ public AtlasPatchManager(AtlasGraph atlasGraph, AtlasTypeRegistry typeRegistry,
         this.context = new PatchContext(atlasGraph, typeRegistry, indexer, entityGraphMapper);
     }
 
-    @PostConstruct
```
ashutoshm, line 51: This approach is fine. However, i would keep the init method in tact and just remove the @PostConstruct annotation.

Call _init_ in the _applyAll_ method.

I verified this change and it is fine.

sarathsubramanian:
Thanks for the details Anshul Mehta. I have a question:

Internal patch attributes like "**__createdBy, __modifiedBy, __timestamp, __modificationTimestamp, __patch.id, __patch.description, __patch.type, __patch.action, __patch.state**" are created in GraphBackedSearchIndexer which is Spring annotated component with Order(1)

The AtlasPatchService has an Order(3), so the internal indexes should be created before Patch Manager is initialized right? Do you see a race scenario where the init() happens during dependency injection of AtlasPatchManager in AtlasPatchService?

PR #143, opened 2021-08-20
Log solr wait searcher conf only if index backend is solr

mehtaanshul:
Log "atlas.graph.index.search.solr.wait-searcher" only when index backend is Solr. 
This PR removes the extra log which gives an error in case we are using elasticsearch as index backend and have not solr config in application properties. Also, the above config is already getting logged at line 350.


mehtaanshul:
@nikhilbonte @nixonrodrigues

PR #152, opened 2021-11-15
Fix NPE when using Elasticsearch

kevinw66:
When using elasticsearch, if the `atlas.graph.index.search.solr.wait-searcher` is not configured, it will produce NullPointerException


ashutoshm (approved):

ctlhsilva:
Please merge the fix.

PR #158, opened 2022-01-17
Simply if clauses in AtlasClient.call_api

dolfinus:
There are a lot if nested if clauses in the `AtlasClient.call_api` method which can be removed or rearranged

### DIFF
--- a/intg/src/main/python/apache_atlas/client/base_client.py
+++ b/intg/src/main/python/apache_atlas/client/base_client.py
@@ -66,7 +66,7 @@ class AtlasClient:
             params['params'] = query_params
 
         if request_obj is not None:
-            params['data'] = json.dumps(request_obj)
+            params['json'] = json.dumps(request_obj)
 
         if LOG.isEnabledFor(logging.DEBUG):
             LOG.debug("------------------------------------------------------")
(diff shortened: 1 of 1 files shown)


dolfinus:
Waiting for review

dolfinus:
@mneethiraj @umeshpadashetty could anyone review, please?

intg/src/main/python/apache_atlas/client/base_client.py
```diff
@@ -68,51 +68,36 @@ def call_api(self, api, response_type=None, query_params=None, request_obj=None)
...
-
-        if response is None:
-            return None
-        elif response.status_code == api.expected_status:
+        log.debug("------------------------------------------------------")
+        log.debug("Call         : %s %s", api.method, path)
+        log.debug("Content-type : %s", api.consumes)
+        log.debug("Accept       : %s", api.produces)
+
+        method = HTTPMethod(api.method)
```
mneethiraj: Is `method` necessary? Why not use `api.method.value` in the following line?

intg/src/main/python/apache_atlas/client/base_client.py
```diff
@@ -68,51 +68,36 @@ def call_api(self, api, response_type=None, query_params=None, request_obj=None)
...
+                if response_type == str:
+                    return json.dumps(response.json())
 
-                    return type_coerce(response.json(), response_type)
-                else:
-                    return None
+                return type_coerce(response.json(), response_type)
             except Exception as e:
-                log.exception("Exception occurred while parsing response with msg: %s", e)
+                log.exception("Exception occurred while parsing response")
```
mneethiraj, line 92: For easier troubleshooting, I suggest to retain including of exception details in the log:
`  log.exception("Exception occurred while parsing response with msg: %s", e)
`

dolfinus: `log.exception` already prints exception stacktrace and message

intg/src/main/python/apache_atlas/client/base_client.py
dolfinus: Fixed

mneethiraj (approved):

PR #181, opened 2022-05-27
fix: remove searchfilter params for TypeDef GET

Anandkarthick:
following TypeDef GET requests don't need params

- /v2/types/typedefs
- /v2/types/typedefs/headers

Change has been tested.


mneethiraj:
@Anandkarthick - thank you for the patch. APIs to retrieve typeDefs and typeDefHeaders do support search parameters, like: type, name, supertype, servicetype, notsupertype, notservicetype, notname. Hence searchFilter parameter in Python method is useful to send necessary filters.

PR #197, opened 2022-09-16
ATLAS-4676: do not expect response body to be a valid JSON

dolfinus:


mneethiraj:
@dolfinus  - thank you reporting this issue and the patch.  Use of f-string formatting will require Python 3.6 or later. I suggest to retain use of .format() method to change Python version requirement. How about the following alternate fix?

```
        body = ""

        if response.content:
            try:
                body = response.json()
            except Exception:
                body = response.content

        msg = "Metadata service API {method} {path} failed with status {status}. Response body is: {body}".format(**{'method': api.method, 'path': api.path, 'status': response.status_code, 'body': body})

```

dolfinus:
Python 3.6 already reached end of life (23.12.2021), 2.7 is not supported since 01.01.2020. A lot of packages already dropped support of such old Python versions, why bother keeping it?

PR #203, opened 2022-10-19
Remove os.path.join causing incorrect windows path

bdsoha:
When running on *Windows* the `os.path.join` using the operating systems `os.sep=\` which causing the full url to be incorrect.

Using `urllib.parse` might be a little bit slower, but is less error-prone.

### DIFF
--- a/intg/src/main/python/apache_atlas/client/base_client.py
+++ b/intg/src/main/python/apache_atlas/client/base_client.py
@@ -18,9 +18,9 @@
 import copy
 import json
 import logging
-import os
 
 from requests import Session
+from urllib.parse import urljoin
 
 from apache_atlas.client.admin import AdminClient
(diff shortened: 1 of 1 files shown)


bdsoha:
@pinal-shah @mandarambawane Any activity on the *Python Client*?

mneethiraj:
@bdsoha  - thank you for the fix. Changes look good. I created [ATLAS-4704](https://issues.apache.org/jira/browse/ATLAS-4704) to track this fix. The fix will be merged shortly.

bdsoha:
@mneethiraj Any chance a new release will published soon?

The latest version of the client [`0.0.11`](https://pypi.org/project/apache-atlas/0.0.11/) was released on *March 15, 2021*.

bdsoha:
@mneethiraj Thanks!

PR #204, opened 2022-10-29
Branch 2.0

ruoyousi:
Here is a problem.

I found: Solr Kerberos is not effective。
I Hope someone can fix it.

Maybe you can remove the two properties in ranger-admin-site.xml.
But it not the best way.

===

FILE： security-admin\src\main\java\org\apache\ranger\solr\SolrUtil.java

    public QueryResponse runQuery(SolrClient solrClient, SolrQuery solrQuery) throws Throwable {
        if (solrQuery != null) {
            try {
                QueryRequest req      = new QueryRequest(solrQuery, METHOD.POST);
                String       username = PropertiesUtil.getProperty("ranger.solr.audit.user");
                String       password = PropertiesUtil.getProperty("ranger.solr.audit.user.password");
                if (username != null && password != null) {  ////// !!!  check by StringUtils.isNotBlank
                    req.setBasicAuthCredentials(username, password);
                }

                return solrMgr.queryToSolr(req);
            } catch (Throwable e) {
                logger.error("Error from Solr server. ", e);
                throw e;
            }
        }
        return null;
    }


mneethiraj:
@ruoyousi  - the description refers to Apache Ranger code. It is not clear what this patch is about. Please add details and reopen if this is about an Apache Atlas issue.

PR #209, opened 2022-11-30
ATLAS-4682: Update maven-frontend-plugin to version 1.12.1, fix build on Mac ARM64

iemejia:
The current master breaks if built on an ARM64 machine, this is due to the frontend-maven-plugin trying to download a node version that is not available on the arm64 architecture. This PR updates the version of the plugin so it resolves the right version.


iemejia:
@mneethiraj Can you PTAL, or help me get someone who can. Thanks!

iemejia:
Rebased. Can someone PTAL

PR #214, opened 2022-12-17
hive-metadata

Coding-ping:
there are errors when I use import-hive.sh in  hook-bin to import hive metadata,and I can not find import-hive.log in logs.But a ${atlas.log.dir} file is built in hook-bin directory


smileyboy2019:
Atals has not been updated in the past two years, is it not maintained
Atals has not added the latest computing engines such as Flink and Doris, as well as a database that integrates lakes and warehouses

PR #231, opened 2023-06-02
[ATLAS-4759] Do not suppress HTTP 503 error in Python client

dolfinus:
This kind of error should never be suppressed


dolfinus:
@mneethiraj @umeshpadashetty could anyone review, please?

mneethiraj (approved):

mneethiraj:
@dolfinus - thank you for the fix. Fix is merged in master and branch-2.0 branches.

PR #240, opened 2023-10-30
[WIP] ATLAS-4789 adds couchbase hook

chedim:
This adds a Couchbase hook that connects to a Couchbase cluster via DCP and sends cluster metadata onto an Atlas instance using REST APIs.

This is not ready for review, but need to create a PR to check if the SLA have been signed by Couchbase yet.


chedim:
TH

PR #267, opened 2025-01-17
ATLAS-4981: Enable CI for Apache Atlas via GitHub Actions

kumaab:
Enable atlas builds to run via GitHub Actions with Apache Infra.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,78 +33,10 @@ env:
   MAVEN_OPTS: -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false -Dmaven.wagon.http.retryHandler.class=standard -Dmaven.wagon.http.retryHandler.count=3
 
 jobs:
-  build-8:
-    runs-on: ubuntu-22.04
-    timeout-minutes: 60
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,172 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Download build-8 artifacts
+        uses: actions/download-artifact@v4
+        with:
+          name: target-8
+
+      - name: Copy artifacts for docker build
+        run: |
+          cp apache-atlas-*.tar.gz dev-support/atlas-docker/dist
```
mneethiraj: Docker scripts setup Atlas with HBase and Solr, which requires build using profile `dist,external-hbase-solr`. See dev-support/atlas-docker/.env.

How about replacing build-8 and build-11 with steps to build using docker (docker-compose.atlas-build.yml)?

kumaab: I see.

Sure, in this case build with jdk 11 will be skipped. Updated the PR, thanks.

mneethiraj (approved):

PR #291, opened 2025-02-15
ATLAS-4991: Support running IT tests while building Atlas through docker

pinal-shah:

change mvn package to install, to run integration tests while building Atlas through docker

## How was this patch tested?

manually


kumaab:
How about `mvn verify` ?

PR #300, opened 2025-03-04
ATLAS-4995: When user doesn't have permission on one glossary , /glos…

sheetalshah1007:
…sary call throws exception , blocking authorization for all glossaries

Added parameter `skipFailedEntities=true` to glossary (GET) api URL for listing glossaries. This flag will skip any glossaries not having adequate access permissions, fairly returning other entries. 

## How was this patch tested?

Manual testing


farhank31 (approved):
LGTM 👍

sheetalshah1007:
With current pagination design to list glossaries, full page is not utilised in case if there are any skipped glossaries resulting in page-size lesser than given limit.

PR #304, opened 2025-03-07
ATLAS-4995: When user doesn't have permission on one glossary , /glos…

sheetalshah1007:
…sary call throws exception , blocking authorization for all glossaries

This patch handles missing glossaries while ensuring fully paginated results

`@Test : testGetGlossaries_WithPaginationHandlingSkippedGlossaries` validates the pagination behavior of the `getGlossaries() `method while handling skipped glossaries.

`@DataProvider : getAllGlossaryForPaginationDataProvider() `dynamically supplies test cases to the test method. It returns an array of test parameters, where each row represents a different test case.

Reading sample test cases from Data provider :
    `{10, 0, SortOrder.ASCENDING, 10, listAllGuids, null, 1},  `_// Fetch all, no skipping, 1 page_
   ` {5, 2, SortOrder.ASCENDING, 5, listAllGuids, Arrays.asList("guid-3", "guid-7"), 2}, ` // _Fetch 5 starting from index 2, Skip 2 glossaries, expect 2 pages_
    `{10, 6, SortOrder.ASCENDING, 4, listAllGuids, null, 1},  `// _Fetch 10 starting from index 6, only 4 available, 1 page_

## How was this patch tested?

Manual testing

### DIFF
--- a/repository/src/main/java/org/apache/atlas/glossary/GlossaryService.java
+++ b/repository/src/main/java/org/apache/atlas/glossary/GlossaryService.java
@@ -18,6 +18,7 @@
 package org.apache.atlas.glossary;
 
 import org.apache.atlas.AtlasErrorCode;
+import org.apache.atlas.RequestContext;
 import org.apache.atlas.SortOrder;
 import org.apache.atlas.annotation.GraphTransaction;
 import org.apache.atlas.bulkimport.BulkImportResponse;
@@ -71,8 +72,7 @@ import static org.apache.atlas.glossary.GlossaryUtils.getGlossarySkeleton;
 
 @Service
 public class GlossaryService {
-    private static final Logger LOG = LoggerFactory.getLogger(GlossaryService.class);
-
+    private static final Logger LOG                                   = LoggerFactory.getLogger(GlossaryService.class);
     private static final String ATLAS_GLOSSARY_TERM                   = "AtlasGlossaryTerm";
     private static final String NAME_ATTR                             = "name";
     private static final String QUALIFIED_NAME_ATTR                   = "qualifiedName";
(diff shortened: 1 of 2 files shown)


repository/src/main/java/org/apache/atlas/glossary/GlossaryService.java
```diff
@@ -119,30 +121,62 @@ public static boolean isNameInvalid(String name) {
...
-            glossaryEntities = dataAccess.getAtlasEntityStore().getByIds(guidsToLoad, true, false);
-            ret              = new ArrayList<>();
+        // Enable skipping failed entities during processing
+        RequestContext.get().setSkipFailedEntities(true);
 
-            for (AtlasEntity glossaryEntity : glossaryEntities.getEntities()) {
-                AtlasGlossary glossary = glossaryDTO.from(glossaryEntity);
+        while (ret.size() < adjustedLimit && currentOffset < glossaryGuids.size()) {
+            // Fetch next batch of GUIDs
+            PaginationHelper<String> paginationHelper = new PaginationHelper<>(glossaryGuids, currentOffset, adjustedLimit - ret.size());
```
jayendrap: can we initialize the PaginationHelper only once and have method in that to return the list of guids based on offset and adjusted limit?

sheetalshah1007: yes, recent patches include this change

repository/src/test/java/org/apache/atlas/glossary/GlossaryServiceTest.java
```diff
@@ -1270,6 +1300,90 @@ public void testIncorrectFileException() {
...
+
+        mockedEntityStore = mock(AtlasEntityStore.class);
+        when(mockedDataAccess.getAtlasEntityStore()).thenReturn(mockedEntityStore);
+
+        List<String> finalGuidsToSkip = (guidsToSkip == null) ? Collections.emptyList() : guidsToSkip;
+
+        List<String> guidsToReturn = new ArrayList<>(allGlossaryGuids);
+        guidsToReturn.removeAll(finalGuidsToSkip);
+
+        try (MockedStatic<AtlasGraphUtilsV2> mockedGraphUtilsClass = Mockito.mockStatic(AtlasGraphUtilsV2.class)) {
```
jayendrap, line 1333: If an exception is thrown then we should fail the test.

repository/src/test/java/org/apache/atlas/glossary/GlossaryServiceTest.java
```diff
@@ -1270,6 +1300,90 @@ public void testIncorrectFileException() {
...
+
+            // Mocking the retrieval of glossary GUIDs from the system
+            mockedGraphUtilsClass.when(() -> AtlasGraphUtilsV2.findEntityGUIDsByType(anyString(), any()))
+                    .thenReturn(allGlossaryGuids);
+
+            // Create a list of AtlasEntity and AtlasGlossary objects
+            for (String guid : guidsToReturn) {
```
jayendrap: Since on line 1363, we are creating already creating the list of AtlasEntity objects to be returned, then do we need this block?

repository/src/main/java/org/apache/atlas/glossary/GlossaryService.java
```diff
@@ -1237,6 +1273,16 @@ List<T> getPaginatedList() {
             return ret;
         }
 
+        List<T> getPaginatedList(int offset, int limit) {
+            if (offset >= maxSize) {
+                return Collections.emptyList();
+            }
+
+            int pageStart = Math.max(0, offset);
```
jayendrap: pageStart isn't used, so we can remove this. Also pageStart and pageEnd variables will conflict with class level variable names

repository/src/main/java/org/apache/atlas/glossary/GlossaryService.java
```diff
@@ -121,26 +121,56 @@ public static boolean isNameInvalid(String name) {
...
+        int                 maxSize       = glossaryGuids.size();
+
+        // If limit is negative, use maxSize; otherwise, take the minimum of limit and maxSize
+        int adjustedLimit = (limit < 0) ? maxSize : Integer.min(maxSize, limit);
 
-        List<AtlasGlossary>                  ret;
-        List<String>                         guidsToLoad = paginationHelper.getPaginatedList();
-        AtlasEntity.AtlasEntitiesWithExtInfo glossaryEntities;
+        // Enable skipping failed entities during processing
+        RequestContext.get().setSkipFailedEntities(true);
```
jayendrap: Put the code till line 172 in try block and in finally block unset skipFailedEntities, else just in case there is an exception the skipFailedEntities will remain true.

repository/src/test/java/org/apache/atlas/glossary/GlossaryServiceTest.java
sheetalshah1007: added catch block for this

repository/src/test/java/org/apache/atlas/glossary/GlossaryServiceTest.java
sheetalshah1007: removed this block; glossaryDTO is mocked while mocking getByIds()

repository/src/main/java/org/apache/atlas/glossary/GlossaryService.java
sheetalshah1007: removed pageStart usage and pageEnd is renamed to `localPageEnd`

repository/src/main/java/org/apache/atlas/glossary/GlossaryService.java
sheetalshah1007: Done

jayendrap (approved):

pinal-shah (approved):

PR #313, opened 2025-03-26
ATLAS-5005 : Basic search entity filter validation

aditya-gupta36:

Validation for FilterCriteria parameter was absence while API call.

## How was this patch tested?
unit tests, manual tests, API curl command testing done, Raised the patch for pre-commit build.

### DIFF
--- a/intg/src/main/java/org/apache/atlas/AtlasErrorCode.java
+++ b/intg/src/main/java/org/apache/atlas/AtlasErrorCode.java
@@ -179,6 +179,8 @@ public enum AtlasErrorCode {
     LINEAGE_ON_DEMAND_NOT_ENABLED(400, "ATLAS-400-00-100", "Lineage on demand config: {0} is not enabled"),
     INVALID_TOPIC_NAME(400, "ATLAS-400-00-101", "Unsupported topic name : {0}"),
     UNSUPPORTED_TYPE_NAME(400, "ATLAS-400-00-102", "Unsupported {0} name. Names such as 'description','version','options','name','servicetype' are not supported"),
+    INVALID_OPERATOR_PASSED(400, "ATLAS-400-00-103", "Invalid Operator Passed: for attribute name: {0}"),
+    BLANK_ATTRIBUTES(400, "ATLAS-400-00-104", "Attribute Name and Attribute Value can't be empty!"),
 
     UNAUTHORIZED_ACCESS(403, "ATLAS-403-00-001", "{0} is not authorized to perform {1}"),
 
--- a/webapp/src/test/java/org/apache/atlas/web/integration/BasicSearchIT.java
+++ b/webapp/src/test/java/org/apache/atlas/web/integration/BasicSearchIT.java
@@ -138,6 +138,11 @@ public class BasicSearchIT extends BaseResourceIT {
                 LOG.info("TestDescription  :{}", testExpectation.testDescription);
                 LOG.info("SearchParameters :{}", testExpectation.searchParameters);
 
+                SearchParameters parameters = testExpectation.getSearchParameters();
+                if (parameters.getEntityFilters() == null || parameters.getEntityFilters().getAttributeName() == null) {
+                    continue;
+                }
+
                 AtlasSearchResult searchResult = atlasClientV2.facetedSearch(testExpectation.searchParameters);
 
                 if (testExpectation.expectedCount > 0) {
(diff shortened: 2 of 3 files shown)


intg/src/main/java/org/apache/atlas/AtlasErrorCode.java
```diff
@@ -179,6 +179,8 @@ public enum AtlasErrorCode {
     LINEAGE_ON_DEMAND_NOT_ENABLED(400, "ATLAS-400-00-100", "Lineage on demand config: {0} is not enabled"),
     INVALID_TOPIC_NAME(400, "ATLAS-400-00-101", "Unsupported topic name : {0}"),
     UNSUPPORTED_TYPE_NAME(400, "ATLAS-400-00-102", "Unsupported {0} name. Names such as 'description','version','options','name','servicetype' are not supported"),
+    INVALID_OPERATOR_PASSED(400, "ATLAS-400-00-103", "Invalid Operator Passed: for attribute name: {0}"),
```
chaitalicod: Can you rename this to INVALID_OPERATOR?

webapp/src/test/java/org/apache/atlas/web/integration/BasicSearchIT.java
```diff
@@ -138,6 +138,11 @@ public void testDiscoveryWithSearchParameters(String jsonFile) {
                 LOG.info("TestDescription  :{}", testExpectation.testDescription);
                 LOG.info("SearchParameters :{}", testExpectation.searchParameters);
 
+                SearchParameters parameters = testExpectation.getSearchParameters();
+                if (parameters.getEntityFilters() == null || parameters.getEntityFilters().getAttributeName() == null) {
+                    continue;
+                }
```
chaitalicod, line 160: This check is good but Can you confirm with a usecase if this is getting checked or addressing invalid values?

webapp/src/test/java/org/apache/atlas/web/integration/BasicSearchIT.java
```diff
@@ -224,6 +245,50 @@ public void testSavedSearch(String jsonFile) {
...
 
+    @Test(dataProvider = "attributeSearchJSONNamesInvalidOperator")
+    public void testAttributeSearchInvalidOperator(String jsonFile) {
+        try {
+            BasicSearchParametersWithExpectation[] testExpectations = TestResourceFileUtils.readObjectFromJson(jsonFile, BasicSearchParametersWithExpectation[].class);
+            assertNotNull(testExpectations);
+            Arrays
+                    .stream(testExpectations)
+                    .map(testExpectation -> testExpectation.getSearchParameters())
+                    .filter(params -> params.getEntityFilters() != null && params.getEntityFilters().getAttributeName() != null)
```
pareshddevalia: unit test method name is invalidOperator and filter condition is for AttributeName.

intg/src/main/java/org/apache/atlas/AtlasErrorCode.java
aditya-gupta36: Yes, It has been renamed.

webapp/src/test/java/org/apache/atlas/web/integration/BasicSearchIT.java
aditya-gupta36: Above has been modified. You can see the latest change.Have used Java 8 there. 

So no need of:  SearchParameters parameters = testExpectation.getSearchParameters();
                if (parameters.getEntityFilters() == null || parameters.getEntityFilters().getAttributeName() == null) {
                    continue;
                }
                
 So basically, it will pass the TC for invalid attributes and operators. First it will get all search parameters from json(containing invalid values) and hit the API, will throw exception and will thus pass the TC:
 Logic: ry {
                            atlasClientV2.facetedSearch(params);
                        }
                        catch (AtlasServiceException e) {
                            assertTrue(e.getMessage().contains("ATLAS-400-00-103"));
                        }

webapp/src/test/java/org/apache/atlas/web/integration/BasicSearchIT.java
aditya-gupta36: Sure, it has been changed.

for invalid operator -> .filter(params -> params.getEntityFilters() != null && params.getEntityFilters().getOperator() != null)

for invalid attribute:->  .filter(params -> params.getEntityFilters() != null && params.getEntityFilters().getAttributeName() != null)

PR #329, opened 2025-04-15
Add support Ranger Group Policy

RusbNoit:

Added two options to disable upper case and remove prefix ROLE_ to support ActiveDirecroty/LDAP groups in Apache Ranger.
Without this options ATLAS convert groups to ROLE_GROUP_NAME. 

Apache Ranger import user and groups from LDAP/Active Directory in lower case or as is.

atlas.authentication.method.ldap.groupRoleToUpper=false
atlas.authentication.method.ldap.groupRolePrefix=false

https://issues.apache.org/jira/browse/ATLAS-4831

## How was this patch tested?

manual tests

### DIFF
--- a/dashboard/src/views/Administrator/Audits/AdminAuditTable.tsx
+++ b/dashboard/src/views/Administrator/Audits/AdminAuditTable.tsx
@@ -15,7 +15,6 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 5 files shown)


RusbNoit:
Open issue https://issues.apache.org/jira/projects/ATLAS/issues/ATLAS-4831?filter=allopenissues

PR #333, opened 2025-04-17
ATLAS-5005 : Basic search entity filter validation

aditya-gupta36:
Changes: Search Filter Validation and Test Cases written for attribute value, name and operator. Incorporated Data providers properly. 

## How was this patch tested?
unit tests, manual tests, API curl command testing done.

### DIFF
--- a/webapp/src/main/java/org/apache/atlas/web/rest/DiscoveryREST.java
+++ b/webapp/src/main/java/org/apache/atlas/web/rest/DiscoveryREST.java
@@ -885,7 +885,50 @@ public class DiscoveryREST {
             if (StringUtils.isNotEmpty(parameters.getQuery()) && parameters.getQuery().length() > maxFullTextQueryLength) {
                 throw new AtlasBaseException(AtlasErrorCode.INVALID_QUERY_LENGTH, Constants.MAX_FULLTEXT_QUERY_STR_LENGTH);
             }
+
+            validateEntityFilter(parameters);
         }
+    }
+
+    private void validateEntityFilter(SearchParameters parameters) throws AtlasBaseException {
+        FilterCriteria entityFilter = parameters.getEntityFilters();
+
+        if (entityFilter == null) {
+            return;
+        }
+
+        if (entityFilter.getCriterion() != null &&
+                !entityFilter.getCriterion().isEmpty()) {
+            if (entityFilter.getCondition() == null || StringUtils.isEmpty(entityFilter.getCondition().toString())) {
+                throw new AtlasBaseException("Condition (AND/OR) must be specified when using multiple filters.");
+            }
+
+            for (FilterCriteria filterCriteria : entityFilter.getCriterion()) {
+                validateCriteria(filterCriteria);
+            }
+        }
+        else {
+            validateCriteria(entityFilter);
+        }
+    }
+
+    private void validateCriteria(SearchParameters.FilterCriteria criteria) throws AtlasBaseException {
+        if (criteria.getOperator() == null) {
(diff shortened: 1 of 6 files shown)


pareshddevalia (approved):
Changes look good to me. Please add the PC once its passed.

aditya-gupta36:
Ok, Added for PC.
Once done will share the PC results.

webapp/src/main/java/org/apache/atlas/web/rest/DiscoveryREST.java
```diff
@@ -885,7 +885,56 @@ private void validateSearchParameters(SearchParameters parameters) throws AtlasB
...
         }
+
+        if (entityFilter.getCriterion() != null &&
+                !entityFilter.getCriterion().isEmpty()) {
+            if (entityFilter.getCondition() == null || StringUtils.isEmpty(entityFilter.getCondition().toString())) {
+                throw new AtlasBaseException("Condition (AND/OR) must be specified when using multiple filters.");
+            }
+
+            for (FilterCriteria filterCriteria : entityFilter.getCriterion()) {
+                if (filterCriteria.getOperator() == null) {
```
rkundam: It appears to have some redundancy. Consider extracting the verification logic into a separate method. This will help in reusing the snippet for both single FilterCriteria and multiple FilterCriteria scenarios, improving maintainability and readability.

webapp/src/main/java/org/apache/atlas/web/rest/DiscoveryREST.java
```diff
@@ -885,7 +885,56 @@ private void validateSearchParameters(SearchParameters parameters) throws AtlasB
...
+                }
+
+                if (requiresValue(filterCriteria.getOperator()) && StringUtils.isBlank(filterCriteria.getAttributeValue())) {
+                    throw new AtlasBaseException(AtlasErrorCode.BLANK_VALUE_ATTRIBUTE);
+                }
+            }
+        }
+        else {
+            if (entityFilter.getOperator() == null) {
+                throw new AtlasBaseException(AtlasErrorCode.INVALID_OPERATOR, entityFilter.getAttributeName());
```
rkundam: Here it only performs a null check for the operator. However, based on the use case mentioned in the Jira, it seems like additional validation for invalid or unsupported operators is expected. Please ensure those checks are included.

rkundam (changes requested):

webapp/src/main/java/org/apache/atlas/web/rest/DiscoveryREST.java
aditya-gupta36: Hi @rkundam,
Yes, checks for invalid or unsupported operators have been implemented. You can refer to the enum block in the following file:
 
https://github.com/apache/atlas/blame/4b702b07650d0376b535275fa8b739c3c01237ac/intg/src/main/java/org/apache/atlas/model/discovery/SearchParameters.java#L431
 
If an operator is not found in the map, the method returns null, which should be treated as an indication of an invalid or unsupported operator. Since this logic is implemented within an enum, throwing an AtlasBaseException directly is not feasible. Therefore, a null return value effectively signals an unsupported case.
 
We had initially considered handling this condition directly within the enum, but since we needed to propagate an appropriate message, we chose this approach instead.
 
Let me know if this aligns with your expectations or if any further clarification is needed.

aditya-gupta36: Hi @rkundam, as suggested I have Resolved the changes.

rkundam (approved):

PR #337, opened 2025-04-22
ATLAS-5017: Patch to replace the long strings set in spark_process at…

chaitalicod:
…tributes

In this fix i have handled existing data with attributes details and sparkplandescription and removed the data from those to avoid OOM errors

## How was this patch tested?
Sanity testing,Manual testing

### DIFF
--- a/repository/src/main/java/org/apache/atlas/repository/patches/ReplaceHugeSparkProcessAttributesPatch.java
+++ b/repository/src/main/java/org/apache/atlas/repository/patches/ReplaceHugeSparkProcessAttributesPatch.java
@@ -1,0 +1,108 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.atlas.repository.patches;
+
+import org.apache.atlas.AtlasConfiguration;
+import org.apache.atlas.exception.AtlasBaseException;
+import org.apache.atlas.pc.WorkItemManager;
+import org.apache.atlas.repository.Constants;
+import org.apache.atlas.repository.graphdb.AtlasGraph;
+import org.apache.atlas.repository.graphdb.AtlasVertex;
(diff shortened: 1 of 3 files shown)


mneethiraj (approved):

repository/src/main/java/org/apache/atlas/repository/patches/ReplaceHugeSparkProcessAttributesPatch.java
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        @Override
+        protected void prepareForExecution() {
+        }
+
+        @Override
+        public void submitVerticesToUpdate(WorkItemManager manager) {
+            AtlasGraph graph = getGraph();
+            Iterable<Object> iterable = graph.query().has(Constants.ENTITY_TYPE_PROPERTY_KEY, TYPE_NAME_SPARK_PROCESS).vertexIds();
```
pinal-shah, line 80: Will it encounter Out of Memory, if there are large number of spark_process entities and each have huge string?
It loads vertices, before getting vertexIds
Can you use pagination?

repository/src/main/java/org/apache/atlas/repository/patches/ReplaceHugeSparkProcessAttributesPatch.java
```diff
@@ -0,0 +1,110 @@
...
+
+            LOG.info("found {} entities of type {}", count, TYPE_NAME_SPARK_PROCESS);
+        }
+
+        @Override
+        protected void processVertexItem(Long vertexId, AtlasVertex vertex, String typeName, AtlasEntityType entityType) {
+            LOG.debug("processItem(typeName={}, vertexId={})", typeName, vertexId);
+
+            try {
+                AtlasGraphUtilsV2.setEncodedProperty(vertex, entityType.getVertexPropertyName(ATTR_NAME_DETAILS), null);
```
pinal-shah: setEncodedProperty --> will get the property value (huge string) and then set
 
 1. can you check for  removeProperty?
 2. getWrappedElement().property(propertyName, null) --> if it allows setting value as null
 
 if above one of the options works without loading the value then, you have to handle usecase where if the property is not present (you can catch exception)

pinal-shah (changes requested):

repository/src/main/java/org/apache/atlas/repository/patches/ReplaceHugeSparkProcessAttributesPatch.java
chaitalicod: After testing with 2300 spark process and each one loaded with this 2 attributes with 19 mb if data with less memory base 2gb 
i tried checking heap usage spike when it fetches all vertexids in a go 
            Iterable<Object> iterable = graph.query().has(Constants.ENTITY_TYPE_PROPERTY_KEY, TYPE_NAME_SPARK_PROCESS).vertexIds();

It didnot casue OOM issues

repository/src/main/java/org/apache/atlas/repository/patches/ReplaceHugeSparkProcessAttributesPatch.java
chaitalicod: I have implemented removeproperty approach in latest patch

pinal-shah (approved):

PR #339, opened 2025-04-28
ATLAS-5025: Skip setLabel call when labels are empty

Amruth-github:
ATLAS-5025: Skip setLabel call when labels are empty

The setLabel during notification processing introduces a overhead while ingesting a message as it involves a call to janus graph even though we don't have any labels to set.

We can make this call conditional as this won't be required if we don't have any labels to set.

Tests planned:
1. PC build passed: https://ci-builds.apache.org/job/Atlas/job/PreCommit-ATLAS-Build-Test/1855/console
2. Manual Tests
3. Functional Tests

### DIFF
--- a/repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
+++ b/repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
@@ -336,7 +336,7 @@ public class EntityGraphMapper {
 
         setCustomAttributes(ret, entity);
 
-        if (RequestContext.getIsUpdateNotification() || CollectionUtils.isNotEmpty(entity.getLabels())) {
+        if (RequestContext.get().getIsUpdateNotification() || CollectionUtils.isNotEmpty(entity.getLabels())) {
             setLabels(ret, entity.getLabels());
         }
 
--- a/repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
+++ b/repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
@@ -336,7 +336,7 @@ public class EntityGraphMapper {
 
         setCustomAttributes(ret, entity);
 
-        if (RequestContext.get().getIsUpdateNotification() || CollectionUtils.isNotEmpty(entity.getLabels())) {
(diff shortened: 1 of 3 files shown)


server-api/src/main/java/org/apache/atlas/RequestContext.java
```diff
@@ -44,8 +44,9 @@
 public class RequestContext {
     private static final Logger METRICS = LoggerFactory.getLogger("METRICS");
 
-    private static final ThreadLocal<RequestContext> CURRENT_CONTEXT    = new ThreadLocal<>();
-    private static final Set<RequestContext>         ACTIVE_REQUESTS    = new HashSet<>();
+    private static final ThreadLocal<RequestContext> CURRENT_CONTEXT           = new ThreadLocal<>();
+    private static final Set<RequestContext>         ACTIVE_REQUESTS           = new HashSet<>();
+    private static final ThreadLocal<Boolean>        isUpdateNotification      = new ThreadLocal<>();
```
mneethiraj: Instead of having `isUpdateNotification` as a `static ThreadLocal<>`, have it as a instance member of `RequestContext` (similar to other members like `requestTime`). This will also make it unnecessary to reset the flag in NotificationHookConsumer.

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
```diff
@@ -336,7 +336,7 @@ public AtlasVertex createVertexWithGuid(AtlasEntity entity, String guid) throws
 
         setCustomAttributes(ret, entity);
 
-        if (RequestContext.getIsUpdateNotification() || CollectionUtils.isNotEmpty(entity.getLabels())) {
+        if (RequestContext.get().getIsUpdateNotification() || CollectionUtils.isNotEmpty(entity.getLabels())) {
```
mneethiraj: Since the vertext is being created in this method (`createVertexWithGuid()`), is it necessary to check whether the notification is of type update? Also, note that it is possible for an entity to be created even while processing notifications of type `ENTITY_FULL_UPDATE_V2` or `ENTITY_PARTIAL_UPDATE_V2 `.

Only checking if `entity.getLabels()` is not empty before calling `setLabels()` should be enough. Please review and update.

mneethiraj (approved):

PR #356, opened 2025-05-22
ATLAS-5032: Fix basic search when querying by long attribute values

fsi-yuvraj:
JIRA : https://issues.apache.org/jira/browse/ATLAS-5032

### Background:
When performing a search using long entity names with attributes like qualifiedName and the StartsWith operator, results are not returned as expected.

### Root Cause:
The qualifiedName attribute is an indexed key. However, Solr's default standard tokenizer has a maximum token length of 255 characters. When entity names exceed this length, the tokenizer fails to parse the value correctly, leading to search failures.

#### Approach 1 : Update Solr Configuration to Increase Max Token Length
- Modify the Solr schema to increase the maxTokenLength option.
- This allows Solr to properly tokenize field values as per set length.
- Impact:
-- Requires full reindexing of all existing data to apply the new schema.
  
#### [Existing PR] Approach 2 : Approach 2: Handle Long Value Search by Querying JanusGraph
- For long value searches on indexed keys, query to janusgraph instead of solr
- Impact:
-- This will affect performance as query is executed at janus

## How was this patch tested?
- PC : https://ci-builds.apache.org/job/Atlas/job/PreCommit-ATLAS-Build-Test/1894
- Added unit tests to validate the logic.
- Manually tested the scenario described in the JIRA issue to ensure accurate search results with long entity names.

### DIFF
--- a/common/src/main/java/org/apache/atlas/repository/Constants.java
+++ b/common/src/main/java/org/apache/atlas/repository/Constants.java
@@ -152,6 +152,7 @@ public final class Constants {
     public static final String INDEX_SEARCH_MAX_RESULT_SET_SIZE        = "atlas.graph.index.search.max-result-set-size";
     public static final String INDEX_SEARCH_TYPES_MAX_QUERY_STR_LENGTH = "atlas.graph.index.search.types.max-query-str-length";
     public static final String INDEX_SEARCH_TAGS_MAX_QUERY_STR_LENGTH  = "atlas.graph.index.search.tags.max-query-str-length";
+    public static final String INDEX_SEARCH_SOLR_MAX_TOKEN_LENGTH      = "atlas.graph.solr.index.search.max-token-length";
     public static final String INDEX_SEARCH_VERTEX_PREFIX_PROPERTY     = "atlas.graph.index.search.vertex.prefix";
     public static final String INDEX_SEARCH_VERTEX_PREFIX_DEFAULT      = "$v$";
     public static final String MAX_FULLTEXT_QUERY_STR_LENGTH = "atlas.graph.fulltext-max-query-str-length";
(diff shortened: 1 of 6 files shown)


repository/src/main/java/org/apache/atlas/discovery/SearchProcessor.java
```diff
@@ -671,6 +672,10 @@ protected AtlasGraphQuery toGraphFilterQuery(Set<? extends AtlasStructType> stru
 
                     for (AtlasStructType structType : structTypes) {
                         String qualifiedName = structType.getVertexPropertyName(criteria.getAttributeName());
+                        if (isIndexSearchable(criteria, structType)) {
```
chaitalicod, line 675: This fix was tested for how many hive_columns as for a hive_table for multiple hive_tables with this huge qualifiedname and name using basic search ?

common/src/main/java/org/apache/atlas/repository/Constants.java
```diff
@@ -152,6 +152,7 @@ public final class Constants {
     public static final String INDEX_SEARCH_MAX_RESULT_SET_SIZE        = "atlas.graph.index.search.max-result-set-size";
     public static final String INDEX_SEARCH_TYPES_MAX_QUERY_STR_LENGTH = "atlas.graph.index.search.types.max-query-str-length";
     public static final String INDEX_SEARCH_TAGS_MAX_QUERY_STR_LENGTH  = "atlas.graph.index.search.tags.max-query-str-length";
+    public static final String INDEX_SEARCH_SOLR_MAX_TOKEN_LENGTH      = "atlas.graph.solr.index.search.max-token-length";
```
aditya-gupta36, line 155: Adds a new configurable constant tied to Solr’s default maxTokenLen (255).

Useful for operators (STARTS_WITH, ENDS_WITH, CONTAINS) that can’t safely be processed via the index when values exceed the token length.

Why this helps: avoids Solr truncation and allows search to fall back to the graph safely on long values.

repository/src/main/java/org/apache/atlas/discovery/SearchProcessor.java
```diff
@@ -866,6 +871,10 @@ private boolean isIndexSearchable(FilterCriteria filterCriteria, AtlasStructType
                 } else if (operator == SearchParameters.Operator.CONTAINS && AtlasAttribute.hastokenizeChar(attributeValue) && indexType == null) { // indexType = TEXT
                     LOG.debug("{} operator found for string (TEXT) attribute {} and special characters found in filter value {}, deferring to in-memory or graph query (might cause poor performance)", operator, qualifiedName, attributeValue);
 
+                    ret = false;
+                } else if ((operator == SearchParameters.Operator.STARTS_WITH || operator == SearchParameters.Operator.ENDS_WITH || operator == SearchParameters.Operator.CONTAINS)  && attributeValue.length() > SOLR_MAX_TOKEN_STR_LENGTH) {
```
aditya-gupta36, line 875: Bypasses index for these operators when value length exceeds the limit.

Falls back to a graph traversal!

aditya-gupta36 (approved):
The changes looks good to me, falls back to graph queries when index-based search isn’t safe due to length limits of qualifiedName

fsi-yuvraj:
Similar Changes merged by another user in https://github.com/apache/atlas/pull/650

PR #360, opened 2025-05-23
ATLAS-5043 : import-hive.sh is failing when shell entity id present during import

VinayakMarraiya230920:

(Please fill in changes proposed in this fix. Create an issue in ASF JIRA before opening a pull request and
set the title of the pull request which starts with
the corresponding JIRA issue number. (e.g. ATLAS-XXXX: Fix a typo in YYY))

## How was this patch tested?

(Please explain how this patch was tested. Ex: unit tests, manual tests)
(If this patch involves UI changes, please attach a screenshot; otherwise, remove this)

### DIFF
--- a/addons/hive-bridge/src/main/java/org/apache/atlas/hive/bridge/HiveMetaStoreBridge.java
+++ b/addons/hive-bridge/src/main/java/org/apache/atlas/hive/bridge/HiveMetaStoreBridge.java
@@ -711,6 +711,9 @@ public class HiveMetaStoreBridge {
     private AtlasEntityWithExtInfo toTableEntity(AtlasEntity database, final Table hiveTable, AtlasEntityWithExtInfo table) throws AtlasHookException {
         if (table == null) {
             table = new AtlasEntityWithExtInfo(new AtlasEntity(HiveDataTypes.HIVE_TABLE.getName()));
+        } else if (table.getEntity().getAttribute(ATTRIBUTE_NAME) == null) {
+            table.getEntity().setAttribute(ATTRIBUTE_NAME, hiveTable.getTableName().toLowerCase());
+            table.getEntity().setIsIncomplete(false);
         }
 
         AtlasEntity tableEntity        = table.getEntity();


addons/hive-bridge/src/main/java/org/apache/atlas/hive/bridge/HiveMetaStoreBridge.java
```diff
@@ -711,6 +711,9 @@ private AtlasEntityWithExtInfo toTableEntity(AtlasEntity database, Table hiveTab
     private AtlasEntityWithExtInfo toTableEntity(AtlasEntity database, final Table hiveTable, AtlasEntityWithExtInfo table) throws AtlasHookException {
         if (table == null) {
             table = new AtlasEntityWithExtInfo(new AtlasEntity(HiveDataTypes.HIVE_TABLE.getName()));
+        } else if (table.getEntity().getAttribute(ATTRIBUTE_NAME) == null) {
+            table.getEntity().setAttribute(ATTRIBUTE_NAME, hiveTable.getTableName().toLowerCase());
+            table.getEntity().setIsIncomplete(false);
```
mneethiraj: Why is it necessary to set `isInComplete = false` here?

VinayakMarraiya230920: When Hive tables are created, Atlas might temporarily create shell entities—incomplete
If these shell entities are included in the export file before they are fully resolved, the import will fail with an error like:
ERROR - Invalid instance creation/updation parameters passed: hive_table.name: mandatory
However, when we run  import-hive in Atlas, it completes these shell entities by setting isInComplete = false and filling in the missing details.

mneethiraj: ok. Shouldn't `isIncomplete` be set to `false` for all tables being imported, i.e. not just for the tables that don't have attribute `name`?  Consider moving line 716 outside the if/else block.

VinayakMarraiya230920: You're right — thanks for pointing that out. I've updated the code so that isIncomplete is now set to false for all imported tables, not just the ones missing the attribute. The call to setIsIncomplete(false) has been moved outside the if/else block to ensure it's always applied.

mneethiraj (approved):

PR #373, opened 2025-06-05
ATLAS-5053: Add recursive validation in Basic Search API to fix regression issue

aditya-gupta36:
- Initially under basic and classification search, only basic [Add filter] option on UI,  validation was done. Also, for [Add Group FIlter]option on UI  it was throwing exception and errors.
- So in this PR we have fixed that issue and we have handled filter search validation for both basic and complex filter groups.
- Group FIlter - it recursively calls Filter SearchParameters. eg: In FilterCriteria class we have List<FilterCriteria> criterion, so entityFilters attribute in SearchParametrs class recursively calls criterion in itself multiple times.
- Apart have added TC to test negative scenarios for quick search and basic search

Conclusion:

  "entityFilters": {
>     "condition": "AND",
>     "criterion": [
>       {
>         "attributeName": "__classificationNames",
>         "operator": "eq",
>         "attributeValue": "GameCatA"
>       },
>       {
>         "condition": "AND",
>         "criterion": [
>           {
>             "attributeName": "year_held",
>             "operator": "eq",
>             "attributeValue": "2006"
>           }
>         ]
>       }
>     ]
>   },
>   "tagFilters": null,
>   "attributes": [
>     "__classificationNames",
>     "__typeName",
>     "year_held"
>   ],
>   "limit": 25,
>   "offset": 0,
>   "typeName": "tournament",
>   "classification": null,
>   "termName": null
> }

- As one can see, inside entity filter, we have criterion which in itself has condition and criterion i.e following nested pattern[recursively filtered], si in such case it gives: 
{
    "errorCode": "ATLAS-400-00-103",
    "errorMessage": "Invalid operator specified for attribute: null"
}

So we have fixed this Complex filter validation[i.e. Add Filter Group Option]. Earlier it was only able to handle basic filter[i.e. normal Add Filter Option]
 
## How was this patch tested?
- Curl Request
- Postman Request
- UI testing
- PC: https://ci-builds.apache.org/job/Atlas/job/PreCommit-ATLAS-Build-Test/1908/
- CI Build

### DIFF
--- a/webapp/src/main/java/org/apache/atlas/web/rest/DiscoveryREST.java
+++ b/webapp/src/main/java/org/apache/atlas/web/rest/DiscoveryREST.java
@@ -891,28 +891,43 @@ public class DiscoveryREST {
     }
 
     private void validateEntityFilter(SearchParameters parameters) throws AtlasBaseException {
-        FilterCriteria entityFilter = parameters.getEntityFilters();
+        validateNestedCriteria(parameters.getEntityFilters());
+        validateNestedCriteria(parameters.getTagFilters());
+    }
 
-        if (entityFilter == null) {
+    private void validateNestedCriteria(SearchParameters.FilterCriteria criteria) throws AtlasBaseException {
+        if (criteria == null) {
+            return; // Nothing to validate
+        }
+
+        boolean hasComposite = criteria.getCriterion() != null && !criteria.getCriterion().isEmpty();
+        boolean hasLeaf = StringUtils.isNotEmpty(criteria.getAttributeName())
+                || criteria.getOperator() != null
+                || StringUtils.isNotEmpty(criteria.getAttributeValue());
+
+        if (!hasComposite && !hasLeaf) {
+            // It's an empty filter object — skip (backward compatibility)
             return;
         }
 
-        if (entityFilter.getCriterion() != null &&
-                !entityFilter.getCriterion().isEmpty()) {
-            if (entityFilter.getCondition() == null || StringUtils.isEmpty(entityFilter.getCondition().toString())) {
+        if (hasComposite) {
+            if (criteria.getCondition() == null || StringUtils.isEmpty(criteria.getCondition().toString())) {
                 throw new AtlasBaseException("Condition (AND/OR) must be specified when using multiple filters.");
             }
 
-            for (FilterCriteria filterCriteria : entityFilter.getCriterion()) {
-                validateCriteria(filterCriteria);
+            for (FilterCriteria filterCriteria : criteria.getCriterion()) {
+                if (filterCriteria != null) {
+                    validateNestedCriteria(filterCriteria); // Recursive check
+                }
             }
         }
-        else {
-            validateCriteria(entityFilter);
+
+        if (hasLeaf) {
+            validateLeafFilterCriteria(criteria);
         }
     }
 
-    private void validateCriteria(SearchParameters.FilterCriteria criteria) throws AtlasBaseException {
+    private void validateLeafFilterCriteria(SearchParameters.FilterCriteria criteria) throws AtlasBaseException {
         if (criteria.getOperator() == null) {
             throw new AtlasBaseException(AtlasErrorCode.INVALID_OPERATOR, criteria.getAttributeName());
         }
--- a/webapp/src/test/java/org/apache/atlas/web/integration/BasicSearchIT.java
+++ b/webapp/src/test/java/org/apache/atlas/web/integration/BasicSearchIT.java
@@ -35,6 +35,7 @@ import org.apache.atlas.model.typedef.AtlasClassificationDef;
 import org.apache.atlas.model.typedef.AtlasTypesDef;
 import org.apache.atlas.type.AtlasTypeUtil;
 import org.apache.atlas.utils.TestResourceFileUtils;
+import org.apache.commons.lang.StringUtils;
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
@@ -46,7 +47,6 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
-import java.util.function.Predicate;
 
 import static com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility.NONE;
 import static com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility.PUBLIC_ONLY;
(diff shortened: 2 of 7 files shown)


webapp/src/main/java/org/apache/atlas/web/rest/DiscoveryREST.java
```diff
@@ -892,27 +892,39 @@ private void validateSearchParameters(SearchParameters parameters) throws AtlasB
 
     private void validateEntityFilter(SearchParameters parameters) throws AtlasBaseException {
         FilterCriteria entityFilter = parameters.getEntityFilters();
+        FilterCriteria tagFilter = parameters.getTagFilters();
 
-        if (entityFilter == null) {
-            return;
+        if (entityFilter != null && tagFilter == null) {
```
pareshddevalia: This IF condition is redundant—you're repeating validateNestedCriteria(...) calls based on every combination of null checks.
Can you make simpler and more readable code.

aditya-gupta36: Okay, thanks for the review. Will update it.

 if (entityFilter != null) {
            validateNestedCriteria(entityFilter);
 }
 if (tagFilter != null) {
            validateNestedCriteria(tagFilter);
 }

Done.

pareshddevalia (approved):
LGTM

webapp/src/test/java/org/apache/atlas/web/integration/BasicSearchIT.java
```diff
@@ -228,21 +228,33 @@ public void testSavedSearch(String jsonFile) {
     }
 
     @Test
-    public void testAttributeSearchInvalidOperator() {
-        runNegativeSearchTest("search-parameters/operator", "ATLAS-400-00-103", parameters -> parameters.getEntityFilters() != null && parameters.getEntityFilters().getOperator() != null);
+    public void testFilterInvalidOperator() {
```
pareshddevalia: Please fix the nomenclature of method same as rest other method.

aditya-gupta36: Done.

public void testComplexFilterInvalidOperator() {
------
}

webapp/src/main/java/org/apache/atlas/web/rest/DiscoveryREST.java
```diff
@@ -892,27 +892,48 @@ private void validateSearchParameters(SearchParameters parameters) throws AtlasB
...
+            validateNestedCriteria(tagFilter);
+        }
+    }
+
+    private boolean isNotEmpty(FilterCriteria filter) {
+        if (filter == null) {
+            return false;
+        }
+
+        if (filter.getAttributeName() != null && filter.getOperator() != null && filter.getAttributeValue() != null) {
```
rkundam: attributeName, attributeValue, and operator are part of each FilterCriteria inside criterion.
So, checking them directly at the top level might not make sense — this condition might always be false unless we're supporting flat filters.
Do we have any use case where these fields exist directly under FilterCriteria (outside criterion)? If not, probably safe to remove or revisit.

webapp/src/main/java/org/apache/atlas/web/rest/DiscoveryREST.java
```diff
@@ -892,27 +892,48 @@ private void validateSearchParameters(SearchParameters parameters) throws AtlasB
...
+            return true;
+        }
+
+        return filter.getCriterion() != null && !filter.getCriterion().isEmpty();
+    }
+
+    private void validateNestedCriteria(SearchParameters.FilterCriteria criteria) throws AtlasBaseException {
+        if (criteria.getCriterion() != null &&
```
rkundam: We’re checking condition when nested filters exist, but shouldn’t we also validate attributeName, attributeValue, and operator for every FilterCriteria in the list?
Right now, it looks like we only validate the nested structure and skip actual field-level checks on each node.

Given the above two points, does it make sense to keep isNotEmpty() as a separate method?
We're already going deep into the validation in validateNestedCriteria(), and isNotEmpty() is mostly doing partial checks.
Maybe better to include those checks directly in the validation method to keep all related logic in one place.

aditya-gupta36: You're right that we could technically skip the boolean check helper method and call validateNestedCriteria() directly. However, the helper is intentionally included to guard against edge cases, specifically when clients send an empty object ({}) instead of null for entityFilters or tagFilters.

In such scenarios, the filter object is non-null but contains no attributes (i.e., no criterion, no attributeName, etc.). If we skip the check and pass this directly into validation, it would eventually reach the leaf validation and throw exceptions due to missing fields like attributeName or operator.

So the helper (checkEmptyFilterCriteria()) is designed to detect and skip empty filter objects, improving robustness and avoiding false negatives during validation.

I've also rearranged the code for clarity so the flow is easier to understand and maintain.

Thus,
"This code reflects backward compatibility for older and new input behavior. Older clients might send entityFilters: {} or tagFilters: {}, which are technically valid JSON objects but contain no actual filtering logic. This check ensures the system doesn't break or throw validation exceptions for such old and new cases."

webapp/src/main/java/org/apache/atlas/web/rest/DiscoveryREST.java
aditya-gupta36: //here: correct request
curl --location --request POST 'http://localhost:21000/api/atlas/v2/search/basic' \
--header 'Content-Type: application/json' \
--header 'Authorization: Basic YWRtaW46YWRtaW4=' \
--data-raw '{"excludeDeletedEntities":true,"includeSubClassifications":true,"includeSubTypes":true,"includeClassificationAttributes":true,"entityFilters":{"condition":"AND","criterion":[{"attributeName":"__guid","operator":"eq","attributeValue":"wdsewsafcaews"},{"condition":"AND","criterion":[{"attributeName":"__modificationTimestamp","operator":"eq","attributeValue":"1752141624000"}]}]},"tagFilters":{"condition":"AND","criterion":[{"attributeName":"__timestamp","operator":"eq","attributeValue":"1752141282000"},{"condition":"AND","criterion":[{"attributeName":"__modifiedBy","operator":"eq","attributeValue":"sadw"},{"condition":"AND","criterion":[{"attributeName":"__modificationTimestamp","operator":"eq","attributeValue":"1752141696000"}]}]},{"attributeName":"__timestamp","operator":"eq","attributeValue":"1752141282000"}]},"attributes":["__guid","__modificationTimestamp"],"limit":25,"offset":0,"typeName":"_ALL_ENTITY_TYPES","classification":"_ALL_CLASSIFICATION_TYPES","termName":null}'

//here: negative scenario request. Sending empty object(non null), In this case, it was throwing exception if we dont handle with helper method.
curl --location --request POST 'http://localhost:21000/api/atlas/v2/search/basic' \
--header 'Content-Type: application/json' \
--header 'Authorization: Basic YWRtaW46YWRtaW4=' \
--data-raw '{"excludeDeletedEntities":true,"includeSubClassifications":true,"includeSubTypes":true,"includeClassificationAttributes":true,"entityFilters":{},"tagFilters":{},"attributes":["__guid","__modificationTimestamp"],"limit":25,"offset":0,"typeName":"_ALL_ENTITY_TYPES","classification":"_ALL_CLASSIFICATION_TYPES","termName":null}'

webapp/src/main/java/org/apache/atlas/web/rest/DiscoveryREST.java
```diff
@@ -892,27 +892,47 @@ private void validateSearchParameters(SearchParameters parameters) throws AtlasB
...
 
-        if (entityFilter.getCriterion() != null &&
-                !entityFilter.getCriterion().isEmpty()) {
-            if (entityFilter.getCondition() == null || StringUtils.isEmpty(entityFilter.getCondition().toString())) {
+        if (checkEmptyFilterCriteria(tagFilter)) {
+            validateNestedCriteria(tagFilter);
+        }
+    }
+
+    private boolean checkEmptyFilterCriteria(FilterCriteria criteria) {
```
rkundam: Thanks for the explanation — the reasoning around handling empty {} filters for backward compatibility makes sense.

However, my main concern is that checkEmptyFilterCriteria() is only used at the top level, which means we might miss validating leaf filters that appear deeper in the nested structure.

For example, if a leaf filter is nested inside a criterion list, and it’s not empty but still invalid (like missing an operator), it won’t go through checkEmptyFilterCriteria() — and won’t be validated at all unless it happens to be part of a composite filter that triggers validateNestedCriteria().

What I’m suggesting is:

Instead of calling checkEmptyFilterCriteria() just once at the top,
We should move that logic inside validateNestedCriteria() so every filter — whether it’s top-level or nested — is consistently checked and validated.
This keeps the empty filter check (for {} support), but also ensures all filters with actual content get validated properly.

Code Reference:
    private void validateEntityFilter(SearchParameters parameters) throws AtlasBaseException {
        FilterCriteria entityFilter = parameters.getEntityFilters();
        FilterCriteria tagFilter = parameters.getTagFilters();

        validateNestedCriteria(entityFilter);
        validateNestedCriteria(tagFilter);
    }

    private void validateNestedCriteria(SearchParameters.FilterCriteria criteria) throws AtlasBaseException {
        if (criteria == null) {
            return;
        }

        boolean hasComposite = criteria.getCriterion() != null && !criteria.getCriterion().isEmpty();
        boolean hasLeaf = StringUtils.isNotEmpty(criteria.getAttributeName())
                || criteria.getOperator() != null
                || StringUtils.isNotEmpty(criteria.getAttributeValue());

        if (hasComposite) {
            if (criteria.getCondition() == null || StringUtils.isEmpty(criteria.getCondition().toString())) {
                throw new AtlasBaseException("Condition (AND/OR) must be specified when using multiple filters.");
            }

            for (FilterCriteria filterCriteria : criteria.getCriterion()) {
                if (filterCriteria != null) {
                    validateNestedCriteria(filterCriteria);
                }
            }
        }

        if (hasLeaf) {
            validateLeafFilterCriteria(criteria);
        }
    }

aditya-gupta36: Oh, now I got it.

Thanks for the review and suggestion, Radhika.

I've refactored based on your suggestion. Now the empty filter check is embedded inside validateNestedCriteria(), ensuring that all filters — including nested ones — are either validated or gracefully skipped (if empty). This simplifies the logic, removes the need for checkEmptyFilterCriteria(), and ensures consistent and recursive validation across the tree structure while preserving backward compatibility.
This keeps the empty filter check (for {} support), but also ensures all filters with actual content get validated properly.

aditya-gupta36: Example of Request which substantiates the above changes:

curl 'http://ccycloud-1.fsknox.root.comops.site:31000/api/atlas/v2/search/basic' \
  -X POST \
  -H 'User-Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:140.0) Gecko/20100101 Firefox/140.0' \
  -H 'Accept: application/json, text/javascript, */*; q=0.01' \
  -H 'Accept-Language: en-US,en;q=0.5' \
  -H 'Accept-Encoding: gzip, deflate' \
  -H 'Content-Type: application/json' \
  -H 'X-XSRF-HEADER: EEQgZd00cPbAAJZZr9zA' \
  -H 'X-Requested-With: XMLHttpRequest' \
  -H 'Origin: http://ccycloud-1.fsknox.root.comops.site:31000' \
  -H 'Connection: keep-alive' \
  -H 'Referer: http://ccycloud-1.fsknox.root.comops.site:31000/index.html' \
  -H 'Cookie: SESSION=NDhlYWU2YmYtNTc4OC00NTVlLWI3MmEtNTk1MzY5MjgzZGJl; ATLASSESSIONID=node0p7tci3nip451oc6i1u75z6h7.node0' \
  --data-raw '{
    "excludeDeletedEntities": true,
    "includeSubClassifications": true,
    "includeSubTypes": true,
    "includeClassificationAttributes": true,
    "entityFilters": null,
    "tagFilters": {
      "condition": "AND",
      "criterion": [
        {
          "attributeName": "__timestamp",
          "operator": "eq",
          "attributeValue": "1752219388000"
        },
        {
          "condition": "AND",
          "criterion": []
        }
      ]
    },
    "attributes": [],
    "limit": 25,
    "offset": 0,
    "typeName": null,
    "classification": "_ALL_CLASSIFICATION_TYPES",
    "termName": null
  }' \
  --insecure


Above "criterion" : [] -> intentionally put empty

chaitalicod (approved):

rkundam (approved):

PR #374, opened 2025-06-05
ATLAS-4916 : Improve Efficiency in Access Request Handling by Optimizing Classification Retrieval

VinayakMarraiya230920:
1.Refactored code to reduce the time taken for processing classification updates for large numbers of entities.
2.Optimized internal method calls to avoid redundant operations.

## How was this patch tested?
1.This patch was tested by measuring the time taken by the bulk/classification API to handle 260 items, both before and after the code changes.

2.The test was done using different numbers of classifications (3, 5, 10, and 20). The results were averaged over three runs to ensure accuracy.

Results after code changes:
Classifications	Entities	Time Taken (seconds)
20	                260	          93.48
10	                260	          39.33
5	                260	          25.55
3	                260	          18.07

Results before code changes:
Classifications	Entities	Time Taken (seconds)
20	                  260	157.67
10	                  260	78.84
5	                  260	39.42
3	                  260	23.65

3.The reading above was taken three times and then averaged to get the final result.  
4.Additionally, logs for method-level performance changes were verified in metrics.log.

### DIFF
--- a/authorization/src/main/java/org/apache/atlas/authorize/AtlasAccessRequest.java
+++ b/authorization/src/main/java/org/apache/atlas/authorize/AtlasAccessRequest.java
@@ -17,7 +17,6 @@
  */
 package org.apache.atlas.authorize;
 
-import org.apache.atlas.model.instance.AtlasClassification;
 import org.apache.atlas.model.instance.AtlasEntityHeader;
 import org.apache.atlas.type.AtlasClassificationType;
 import org.apache.atlas.type.AtlasEntityType;
@@ -173,14 +172,10 @@ public class AtlasAccessRequest {
     public Set<String> getClassificationNames(AtlasEntityHeader entity) {
         final Set<String> ret;
 
-        if (entity == null || entity.getClassifications() == null) {
+        if (entity == null || entity.getClassificationNames() == null) {
             ret = Collections.emptySet();
         } else {
-            ret = new HashSet<>();
-
-            for (AtlasClassification classify : entity.getClassifications()) {
-                ret.add(classify.getTypeName());
-            }
+            ret = new HashSet<>(entity.getClassificationNames());
         }
 
         return ret;
--- a/authorization/src/test/java/org/apache/atlas/authorize/simple/AtlasSimpleAuthorizerTest.java
+++ b/authorization/src/test/java/org/apache/atlas/authorize/simple/AtlasSimpleAuthorizerTest.java
@@ -142,7 +142,7 @@ public class AtlasSimpleAuthorizerTest {
         try {
             AtlasEntityHeader entityHeader = new AtlasEntityHeader();
 
-            entityHeader.setClassifications(Arrays.asList(new AtlasClassification("TAG1"), new AtlasClassification("TAG2")));
+            entityHeader.setClassificationNames(Arrays.asList("TAG1", "TAG2"));
 
             AtlasEntityAccessRequest request = new AtlasEntityAccessRequest(null, AtlasPrivilege.ENTITY_ADD_CLASSIFICATION, entityHeader, new AtlasClassification("PII"));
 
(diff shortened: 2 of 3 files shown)


authorization/src/main/java/org/apache/atlas/authorize/AtlasEntityAccessRequest.java
```diff
@@ -76,7 +78,7 @@ public AtlasEntityAccessRequest(AtlasTypeRegistry typeRegistry, AtlasPrivilege a
         this.businessMetadata      = businessMetadata;
         this.attributeName         = attributeName;
         this.typeRegistry          = typeRegistry;
-        this.entityClassifications = super.getClassificationNames(entity);
+        this.entityClassifications = (entity != null && entity.getClassificationNames() != null) ? new HashSet<>(entity.getClassificationNames()) : Collections.emptySet();
```
pinal-shah: this.entityClassifications = CollectionUtils.isNotEmpty(entity.getClassificationNames()) ? new HashSet<>(entity.getClassificationNames()) : Collections.emptySet();

pinal-shah (approved):

authorization/src/main/java/org/apache/atlas/authorize/AtlasEntityAccessRequest.java
```diff
@@ -76,7 +79,7 @@ public AtlasEntityAccessRequest(AtlasTypeRegistry typeRegistry, AtlasPrivilege a
         this.businessMetadata      = businessMetadata;
         this.attributeName         = attributeName;
         this.typeRegistry          = typeRegistry;
-        this.entityClassifications = super.getClassificationNames(entity);
+        this.entityClassifications = (entity != null && CollectionUtils.isNotEmpty(entity.getClassificationNames())) ? new HashSet<>(entity.getClassificationNames()) : Collections.emptySet();
```
mneethiraj: `AtlasAccessRequest.getClassificationNames()` implementation already handles `entity == null` and returns either `Collections.emptySet()` or a new `HashSet` containing classification names. Is the change here necessary?

pinal-shah: This change is necessary, so as to avoid using below
`entityRetriever.toAtlasEntityHeaderWithClassifications
`instead we can make use of
`entityRetriever.toAtlasEntityHeader` wherever possible, which will avoid extra edge traversal call to find classification.

mneethiraj: ok. `AtlasAccessRequest.getClassificationNames()` is called from `AtlasRelationshipAccessRequest` as well. Instead of updating `AtlasEntityAccessRequest`, how about updating `AtlasAccessRequest.getClassificationNames()` to return `entity.getClassificationNames()` (instead of retrieving classifications)?

pinal-shah: yes agreed, we can and can take up in another ticket to update the all possible calling methods to use without classifications.

VinayakMarraiya230920: I've updated the PR based on the comments. Please review it.

authorization/src/main/java/org/apache/atlas/authorize/AtlasAccessRequest.java
```diff
@@ -173,13 +172,13 @@ public String getEntityId(AtlasEntityHeader entity, AtlasTypeRegistry typeRegist
...
 
-        if (entity == null || entity.getClassifications() == null) {
+        if (entity == null || entity.getClassificationNames() == null) {
             ret = Collections.emptySet();
         } else {
             ret = new HashSet<>();
 
-            for (AtlasClassification classify : entity.getClassifications()) {
-                ret.add(classify.getTypeName());
+            for (String classify : entity.getClassificationNames()) {
```
mneethiraj: I suggest replacing lines 178 to 182 with the following:
```
  ret = new HashSet<>(entity.getClassificationNames());
```

VinayakMarraiya230920: Suggested changes are done

PR #394, opened 2025-07-02
ATLAS-5039: Page for TeamList doesn't render Atlas team members in atlas.apache.org

pinal-shah:

Problem: The Apache Atlas Team List in the website is rendered from github pom.xml dynamically.
Since the response header `content-security-policy` from apache atlas website doesn't allow GitHub URL, Team List is not loading

Solution: Team List is fetched in a pre build step and saved in a file and rendered from that file, instead of fetching from github url.
Another Solution is mentioned [here](https://infra.apache.org/tools/csp.html)

## How was this patch tested?

Locally build the doc module and checked

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -5,6 +5,8 @@
   "main": "index.js",
   "scripts": {
     "start": "node ./docz-lib/docz/bin/index.js dev",
+    "fetch-team": "node scripts/fetchTeamList.js",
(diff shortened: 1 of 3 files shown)


mneethiraj (approved):

pawarprasad123 (approved):
LGTM

PR #400, opened 2025-07-15
ATLAS-5070: Startup failure due to NPE in IndexRecoveryService when atlas.index.recovery.start.time is unset

rkundam:

This pull request introduces a null check for the atlas.index.recovery.start.time configuration value, which is explicitly used to trigger index recovery from a specific start time.

If the configuration is not set, the recovery logic is safely skipped without throwing a NullPointerException.
This change ensures that the system can start up cleanly even when atlas.index.recovery.start.time is not configured.

## How was this patch tested?
Tested locally.


mneethiraj (approved):

rkundam:
PC build: https://ci-builds.apache.org/job/Atlas/job/PreCommit-ATLAS-Build-Test/1943/

PR #406, opened 2025-07-19
ATLAS-4920: Atlas Auto Purge Feature

achandel-01:

- This feature provides user configurable cron based cleanup(purging) mechanism for the inactive entities(soft deleted entities).
- Process entities (hive_process,spark_process etc) deletion is introduced.
- A new audit operation, AUTO_PURGE, has been introduced to log automatically purged entities, making them visible in the   Atlas UI for better traceability.
- Historical process entities having no significance can be cleaned.

## How was this patch tested?

- Manual testing (various different scenarios of lineage taken into account)
- Unit testing 

## UI screenshots

auto-purge-ui-1

auto-purge-ui-2

### DIFF
--- a/dashboardv2/public/js/views/audit/AdminAuditTableLayoutView.js
+++ b/dashboardv2/public/js/views/audit/AdminAuditTableLayoutView.js
@@ -244,7 +244,7 @@ define(['require',
                             el.attr('colspan', '8');
                             if (results) {
                                 var adminValues = null;
-                                if (operation == "PURGE") {
+                                if (operation == "PURGE" || operation == "AUTO_PURGE") {
                                     adminText = that.displayPurgeAndImportAudits(auditData);
                                 } else if (operation == "EXPORT" || operation == "IMPORT") {
                                     adminText = that.displayExportAudits(auditData);
@@ -353,7 +353,7 @@ define(['require',
                 var adminValues = '<ul class="col-sm-6">',
                     guids = null,
                     adminTypDetails = Enums.category[obj.operation];
-                if (obj.operation == "PURGE") {
+                if (obj.operation == "PURGE" || operation == "AUTO_PURGE") {
                     guids = obj.results ? obj.results.replace('[', '').replace(']', '').split(',') : guids;
                 } else {
                     guids = obj.model.get('params') ? obj.model.get('params').split(',') : guids;
@@ -430,7 +430,7 @@ define(['require',
                 require(['views/audit/AuditTableLayoutView'], function(AuditTableLayoutView) {
                     var obj = {
                             guid: $(e.target).text(),
-                            titleText: (e.target.dataset.operation == "PURGE") ? "Purged Entity Details: " : "Import Details: "
+                            titleText: (e.target.dataset.operation == "PURGE" || e.target.dataset.operation == "AUTO_PURGE") ? "Purged Entity Details: " : "Import Details: "
                         },
                         modalData = {
                             title: obj.titleText + obj.guid,
--- a/dashboardv3/public/js/views/audit/AdminAuditTableLayoutView.js
+++ b/dashboardv3/public/js/views/audit/AdminAuditTableLayoutView.js
@@ -244,7 +244,7 @@ define(['require',
                             el.attr('colspan', '8');
                             if (results) {
                                 var adminValues = null;
-                                if (operation == "PURGE") {
+                                if (operation == "PURGE" || operation == "AUTO_PURGE") {
                                     adminText = that.displayPurgeAndImportAudits(auditData);
                                 } else if (operation == "EXPORT" || operation == "IMPORT") {
                                     adminText = that.displayExportAudits(auditData);
@@ -353,7 +353,7 @@ define(['require',
                 var adminValues = '<ul class="col-sm-6">',
                     guids = null,
                     adminTypDetails = Enums.category[obj.operation];
-                if (obj.operation == "PURGE") {
+                if (obj.operation == "PURGE" || obj.operation == "AUTO_PURGE") {
                     guids = obj.results ? obj.results.replace('[', '').replace(']', '').split(',') : guids;
                 } else {
                     guids = obj.model.get('params') ? obj.model.get('params').split(',') : guids;
@@ -430,7 +430,7 @@ define(['require',
                 require(['views/audit/AuditTableLayoutView'], function(AuditTableLayoutView) {
                     var obj = {
                             guid: $(e.target).text(),
-                            titleText: (e.target.dataset.operation == "PURGE") ? "Purged Entity Details: " : "Import Details: "
+                            titleText: (e.target.dataset.operation == "PURGE" || e.target.dataset.operation == "AUTO_PURGE") ? "Purged Entity Details: " : "Import Details: "
                         },
                         modalData = {
                             title: obj.titleText + obj.guid,
--- a/repository/src/main/java/org/apache/atlas/repository/store/graph/v1/DeleteHandlerV1.java
+++ b/repository/src/main/java/org/apache/atlas/repository/store/graph/v1/DeleteHandlerV1.java
@@ -57,16 +57,7 @@ import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Set;
-import java.util.Stack;
+import java.util.*;
 
 import static org.apache.atlas.model.TypeCategory.ARRAY;
 import static org.apache.atlas.model.TypeCategory.CLASSIFICATION;
@@ -77,17 +68,8 @@ import static org.apache.atlas.model.instance.AtlasEntity.Status.ACTIVE;
 import static org.apache.atlas.model.instance.AtlasEntity.Status.DELETED;
 import static org.apache.atlas.model.instance.AtlasEntity.Status.PURGED;
 import static org.apache.atlas.model.typedef.AtlasRelationshipDef.PropagateTags.ONE_TO_TWO;
-import static org.apache.atlas.repository.Constants.CLASSIFICATION_EDGE_NAME_PROPERTY_KEY;
-import static org.apache.atlas.repository.Constants.CLASSIFICATION_ENTITY_STATUS;
-import static org.apache.atlas.repository.Constants.CLASSIFICATION_LABEL;
-import static org.apache.atlas.repository.Constants.CLASSIFICATION_NAME_DELIMITER;
-import static org.apache.atlas.repository.Constants.EDGE_PENDING_TASKS_PROPERTY_KEY;
-import static org.apache.atlas.repository.Constants.MODIFICATION_TIMESTAMP_PROPERTY_KEY;
-import static org.apache.atlas.repository.Constants.MODIFIED_BY_KEY;
-import static org.apache.atlas.repository.Constants.PROPAGATED_CLASSIFICATION_NAMES_KEY;
-import static org.apache.atlas.repository.Constants.PROPAGATED_TRAIT_NAMES_PROPERTY_KEY;
-import static org.apache.atlas.repository.Constants.RELATIONSHIPTYPE_TAG_PROPAGATION_KEY;
-import static org.apache.atlas.repository.Constants.RELATIONSHIP_GUID_PROPERTY_KEY;
+import static org.apache.atlas.repository.Constants.*;
+import static org.apache.atlas.repository.Constants.GUID_PROPERTY_KEY;
 import static org.apache.atlas.repository.graph.GraphHelper.getAllClassificationEdges;
 import static org.apache.atlas.repository.graph.GraphHelper.getAssociatedEntityVertex;
 import static org.apache.atlas.repository.graph.GraphHelper.getBlockedClassificationIds;
@@ -121,11 +103,15 @@ import static org.apache.atlas.repository.store.graph.v2.tasks.ClassificationPro
 import static org.apache.atlas.repository.store.graph.v2.tasks.ClassificationPropagateTaskFactory.CLASSIFICATION_PROPAGATION_DELETE;
 import static org.apache.atlas.type.AtlasStructType.AtlasAttribute.AtlasRelationshipEdgeDirection.OUT;
 import static org.apache.atlas.type.Constants.PENDING_TASKS_PROPERTY_KEY;
+import static org.apache.atlas.model.typedef.AtlasBaseTypeDef.ATLAS_TYPE_DATASET;
+import static org.apache.atlas.model.typedef.AtlasBaseTypeDef.ATLAS_TYPE_PROCESS;
 
 public abstract class DeleteHandlerV1 {
     public static final Logger LOG = LoggerFactory.getLogger(DeleteHandlerV1.class);
 
     private static final boolean DEFERRED_ACTION_ENABLED = AtlasConfiguration.TASKS_USE_ENABLED.getBoolean();
+    private static final String  PROCESS_OUTPUTS_EDGE               = "__Process.outputs";
+    private static final String  RELATIONSHIP_ATTRIBUTE_KEY_STRING  = "columnLineages";
 
     protected final GraphHelper graphHelper;
 
@@ -153,12 +139,22 @@ public abstract class DeleteHandlerV1 {
      * @throws AtlasBaseException
      */
     public void deleteEntities(Collection<AtlasVertex> instanceVertices) throws AtlasBaseException {
+        Set<AtlasVertex> deletionCandidateVertices = accumulateDeletionCandidates(instanceVertices);
+        deleteTraitsAndVertices(deletionCandidateVertices);
+    }
+
+    /*
+        accumulate the deletion candidates
+    */
+    public Set<AtlasVertex> accumulateDeletionCandidates(Collection<AtlasVertex> instanceVertices) throws AtlasBaseException {
         final RequestContext   requestContext            = RequestContext.get();
         final Set<AtlasVertex> deletionCandidateVertices = new HashSet<>();
         final boolean          isPurgeRequested          = requestContext.isPurgeRequested();
+        final Set<String>      instanceVertexGuids       = new HashSet<>();
 
         for (AtlasVertex instanceVertex : instanceVertices) {
             final String guid = AtlasGraphUtilsV2.getIdFromVertex(instanceVertex);
+            instanceVertexGuids.add(guid);
 
             if (skipVertexForDelete(instanceVertex)) {
                 if (LOG.isDebugEnabled()) {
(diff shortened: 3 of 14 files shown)


repository/src/main/java/org/apache/atlas/repository/store/graph/v1/DeleteHandlerV1.java
```diff
@@ -77,17 +68,8 @@
...
-import static org.apache.atlas.repository.Constants.CLASSIFICATION_LABEL;
-import static org.apache.atlas.repository.Constants.CLASSIFICATION_NAME_DELIMITER;
-import static org.apache.atlas.repository.Constants.EDGE_PENDING_TASKS_PROPERTY_KEY;
-import static org.apache.atlas.repository.Constants.MODIFICATION_TIMESTAMP_PROPERTY_KEY;
-import static org.apache.atlas.repository.Constants.MODIFIED_BY_KEY;
-import static org.apache.atlas.repository.Constants.PROPAGATED_CLASSIFICATION_NAMES_KEY;
-import static org.apache.atlas.repository.Constants.PROPAGATED_TRAIT_NAMES_PROPERTY_KEY;
-import static org.apache.atlas.repository.Constants.RELATIONSHIPTYPE_TAG_PROPAGATION_KEY;
-import static org.apache.atlas.repository.Constants.RELATIONSHIP_GUID_PROPERTY_KEY;
+import static org.apache.atlas.repository.Constants.*;
```
rkundam: To follow checkstyle coding guidelines, please avoid using wildcard imports (*). Import only the required classes explicitly.

repository/src/main/java/org/apache/atlas/repository/store/graph/v1/DeleteHandlerV1.java
```diff
@@ -57,16 +57,7 @@
...
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Set;
-import java.util.Stack;
+import java.util.*;
```
rkundam: To follow checkstyle coding guidelines, please avoid using wildcard imports (*). Import only the required classes explicitly.

repository/src/main/java/org/apache/atlas/repository/store/graph/v1/DeleteHandlerV1.java
```diff
@@ -183,15 +179,147 @@ public void deleteEntities(Collection<AtlasVertex> instanceVertices) throws Atla
...
+        if (columnLineages != null && !columnLineages.isEmpty()) {
+            AtlasAttribute atlasAttribute= columnLineages.values().iterator().next();
+            String relationshipEdgeLabel = atlasAttribute.getRelationshipEdgeLabel();
+
+            Iterator<AtlasEdge>edgeIterator = GraphHelper.getIncomingEdgesByLabel(process, relationshipEdgeLabel);
+
+            while (edgeIterator.hasNext()) {
+                AtlasVertex columnLineageVertex = edgeIterator.next().getOutVertex();
+                String typeName = columnLineageVertex.getProperty(TYPE_NAME_PROPERTY_KEY, String.class);
+                LOG.info("typeName {} ", typeName);
```
rkundam: Update the comment with more meaningful and descriptive information.

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasEntityStoreV2.java
```diff
@@ -577,6 +577,53 @@ public EntityMutationResponse purgeByIds(Set<String> guids) throws AtlasBaseExce
...
+        EntityMutationResponse response      = new EntityMutationResponse();
+
+        RequestContext requestContext = RequestContext.get();
+        requestContext.setDeleteType(DeleteType.HARD);// hard deleter
+        requestContext.setPurgeRequested(true);
+
+        for (String guid : purgeCandidates) {
+            AtlasVertex vertex = AtlasGraphUtilsV2.findByGuid(graph, guid);
+            if( vertex != null ) {
+                AtlasEntityHeader entityHeader = entityRetriever.toAtlasEntityHeader(vertex);
```
rkundam, line 595: Fetching EntityHeader info for all input GUIDs might be expensive. Did you evaluate whether this is necessary?

repository/src/main/java/org/apache/atlas/services/PurgeService.java
```diff
@@ -0,0 +1,532 @@
...
+                    }
+
+                    String guid = vertex.getProperty(GUID_PROPERTY_KEY, String.class);
+
+                    if (!producedDeletionCandidates.contains(guid)) {
+                        Set<String> instanceVertex = new HashSet<>();
+                        instanceVertex.add(guid);
+
+                        Set<AtlasVertex> deletionCandidates = entityStore.accumulateDeletionCandidates(instanceVertex);
+
```
rkundam, line 204: Right now, accumulateDeletionCandidates is being called once per GUID inside the loop, which may be inefficient. Since accumulateDeletionCandidates supports batch input, we could collect all eligible GUIDs into a set and make a single call outside the loop. This would reduce repeated calls and improve performance. If there’s a specific reason it’s implemented this way, please clarify.

repository/src/main/java/org/apache/atlas/services/PurgeService.java
```diff
@@ -0,0 +1,532 @@
...
+            }
+        }
+
+        @Override
+        protected void processItem(AtlasVertex vertex) {
+            String guid =  vertex.getProperty(GUID_PROPERTY_KEY, String.class);
+            LOG.info("==> processing the entity {}", guid);
+
+            try {
+                if (!isPurgeEnabled && !isEligible(vertex)) {
```
rkundam, line 325: In the condition if (!isPurgeEnabled && !isEligible(vertex)), it seems we only skip processing when both are false. If the intent is to skip when either isPurgeEnabled is false or isEligible(vertex) is false, this should use || instead of &&. Can you confirm the intended logic?

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasEntityStoreV2.java
achandel-01: Yes it is expensive.But we need it show this information to the admin audit side, regardless of what type of entity it is, basic information is shown for every purged guid.Also the existing purge api stores all the entity headers.

repository/src/main/java/org/apache/atlas/services/PurgeService.java
achandel-01: no it is fine, because the code I have written is generic for both , active processes clean up and the pruge jobs.
there are two cases :
1st case isPurgeEnabled = true -->  condition will not work and vertex will get added to the batch and isEligible will not run.
2nd case isPurgeEnabled = false --> isEligble(vertex) will run and isEligible basically checks that process entities are eligible for soft deletion or not.If it is true it will put to the batch.

repository/src/main/java/org/apache/atlas/services/PurgeService.java
achandel-01: it's risky because we don't know , like how many entities in total are there for purge after accumulation. Suppose the fetched entities are tables then we will accumulate its composite as well as upstream entities which can lead to the significant rise in the number of entities.If something goes wrong the whole job will get fail without any results.That chances are very less in doing it individually , as accumulated entities in a single time will not be that large.

webapp/src/main/java/org/apache/atlas/web/resources/AdminResource.java
```diff
@@ -118,15 +119,7 @@
...
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.TimeZone;
+import java.util.*;
```
mandarambawane: Do not import all the classes from package using *
Import only required classes.

repository/src/main/java/org/apache/atlas/services/PurgeService.java
```diff
@@ -0,0 +1,532 @@
...
+import org.apache.atlas.utils.AtlasPerfTracer;
+import org.apache.commons.collections.CollectionUtils;
+import org.apache.commons.configuration.Configuration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.core.annotation.Order;
+import org.springframework.stereotype.Component;
+
+import javax.inject.Inject;
+import java.util.*;
```
mandarambawane: Import only required classes.

repository/src/test/java/org/apache/atlas/services/PurgeServiceTest.java
```diff
@@ -0,0 +1,193 @@
...
+import org.apache.atlas.repository.AtlasTestBase;
+import org.apache.commons.configuration.Configuration;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+import org.testng.Assert;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+import java.lang.reflect.Method;
+import java.util.*;
```
mandarambawane: Import only required classes.

repository/src/main/java/org/apache/atlas/repository/store/graph/v1/DeleteHandlerV1.java
```diff
@@ -77,17 +78,19 @@
...
+import static org.apache.atlas.repository.Constants.RELATIONSHIPTYPE_TAG_PROPAGATION_KEY;
 import static org.apache.atlas.repository.Constants.EDGE_PENDING_TASKS_PROPERTY_KEY;
 import static org.apache.atlas.repository.Constants.MODIFICATION_TIMESTAMP_PROPERTY_KEY;
 import static org.apache.atlas.repository.Constants.MODIFIED_BY_KEY;
-import static org.apache.atlas.repository.Constants.PROPAGATED_CLASSIFICATION_NAMES_KEY;
 import static org.apache.atlas.repository.Constants.PROPAGATED_TRAIT_NAMES_PROPERTY_KEY;
-import static org.apache.atlas.repository.Constants.RELATIONSHIPTYPE_TAG_PROPAGATION_KEY;
-import static org.apache.atlas.repository.Constants.RELATIONSHIP_GUID_PROPERTY_KEY;
+import static org.apache.atlas.repository.Constants.PROPAGATED_CLASSIFICATION_NAMES_KEY;
+import static org.apache.atlas.repository.Constants.CLASSIFICATION_NAME_DELIMITER;
```
rkundam: Organize the imports in alphabetical order to avoid potential Checkstyle violations. Additionally, import only the required classes instead of using wildcard (*) imports for better readability and maintainability.

webapp/src/main/java/org/apache/atlas/web/resources/AdminResource.java
```diff
@@ -162,6 +164,10 @@ public class AdminResource {
     private static final List<String>  TIMEZONE_LIST                    = Arrays.asList(TimeZone.getAvailableIDs());
     private static final String        METRICS_PERSIST_INTERVAL         = "atlas.metrics.persist.schedule";
     private static final String        METRICS_PERSIST_INTERVAL_DEFAULT = "0 0 0/1 * * *";     // 1 hour interval
+    private static final String        PURGE_CRON_EXPRESSION            = "atlas.purge.cron.expression";
+    private static final String        PURGE_CRON_EXPRESSION_DEFAULT    = "* * * 30 2 ?";
```
DishaTalreja3: This cron expression "* * * 30 2 ?" looks invalid because February 30th doesn’t exist. Can you please check if this is right?
Also, it would be good to add a comment about what the schedule is supposed to do.

achandel-01: The requirement is to make it controlled by user , if user wants to enable it or not, so rather than defining the separate flag I did it like it will never run until user enters the valid cron expression. So if user enters cron expression then only these cron jobs will be launched.

repository/src/test/java/org/apache/atlas/services/PurgeServiceTest.java
```diff
@@ -0,0 +1,201 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Random;
+import java.util.Set;
+import java.util.UUID;
+import static org.mockito.Mockito.when;
+
+public class PurgeServiceTest extends AtlasTestBase {
```
mandarambawane: Please write test cases covering next level method calls. By this way test cases will not be redundent.

repository/src/test/java/org/apache/atlas/services/PurgeServiceTest.java
```diff
@@ -0,0 +1,186 @@
...
+        Iterator<AtlasIndexQuery.Result> iterator = Arrays.asList(result).iterator();
+        when(mockIndexQuery.vertices(anyInt(), anyInt())).thenReturn(iterator);
+
+        // mock entityStore to expand deletion candidates to include the same vertex
+        Set<AtlasVertex> deletionCandidates = new HashSet<>();
+        deletionCandidates.add(vertex);
+        when(entityStore.accumulateDeletionCandidates(any())).thenReturn(deletionCandidates);
+
+        // mock WorkItemsQualifier and its results queue
+        PurgeService.WorkItemsQualifier wiq = mock(PurgeService.WorkItemsQualifier.class);
```
chaitalicod: Please follow naming conventions for better understanding in code flow

repository/src/test/java/org/apache/atlas/services/PurgeServiceTest.java
```diff
@@ -0,0 +1,186 @@
...
+        Queue results = new ArrayDeque();
+        AtlasEntityHeader header = new AtlasEntityHeader();
+        header.setGuid("guid-1");
+        header.setTypeName("typeA");
+        results.add(header);
```
chaitalicod: Please consider a Dataprovider or set of data on which you can perform operations while mocking objects

repository/src/test/java/org/apache/atlas/services/PurgeServiceTest.java
```diff
@@ -0,0 +1,186 @@
...
+        verify(entityStore, times(1)).accumulateDeletionCandidates(any());
+
+        // verify that checkProduce got called at least once with a vertex
+        ArgumentCaptor<AtlasVertex> vertexCaptor = ArgumentCaptor.forClass(AtlasVertex.class);
+        verify(wiq, times(1)).checkProduce(vertexCaptor.capture());
+        Assert.assertEquals(vertexCaptor.getValue(), vertex);
+    }
+
+    @Test
+    public void softDeleteProcessEntitiesNoResultsShutsDownWorkersAndUsesCleanupQualifier() throws Exception {
```
chaitalicod: Please add appropriate names for testcases

repository/src/test/java/org/apache/atlas/services/PurgeServiceTest.java
```diff
@@ -0,0 +1,186 @@
...
+        Assert.assertNotNull(response.getPurgedEntities());
+        Assert.assertEquals(response.getPurgedEntities().size(), 1);
+
+        // verify next-level calls
+        verify(purgeService, times(1)).createQualifier(any(), any(), any(), anyInt(), anyInt(), Mockito.eq(true));
+        verify(entityStore, times(1)).accumulateDeletionCandidates(any());
+
+        // verify that checkProduce got called at least once with a vertex
+        ArgumentCaptor<AtlasVertex> vertexCaptor = ArgumentCaptor.forClass(AtlasVertex.class);
+        verify(wiq, times(1)).checkProduce(vertexCaptor.capture());
```
chaitalicod: Why are we checking if checkProduce is getting called here?

chaitalicod (changes requested):

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasEntityStoreV2.java
```diff
@@ -577,6 +577,53 @@ public EntityMutationResponse purgeByIds(Set<String> guids) throws AtlasBaseExce
...
+    @Override
+    @GraphTransaction
+    public EntityMutationResponse schedulePurgeEntities(Set<String>purgeCandidates) throws AtlasBaseException {
+        LOG.info("==> schedulePurgeEntities()");
+
+        Collection<AtlasVertex>purgeVertices = new ArrayList<>();
+        EntityMutationResponse response      = new EntityMutationResponse();
+
+        RequestContext requestContext = RequestContext.get();
+        requestContext.setDeleteType(DeleteType.HARD);// hard deleter
```
chaitalicod: Can you please share how much data this patch was tested with 
what was Atlas memory how much time it took etc

repository/src/test/java/org/apache/atlas/services/PurgeServiceTest.java
achandel-01: It ensures future changes don’t silently drop the critical enqueue step. Without this verify, someone could remove or bypass wiq.checkProduce(vertex) and the test might still pass (because the queue was prefilled), hiding a real bug. By asserting the call and its exact argument, the test fails immediately if enqueueing is removed, reordered, or altered.

repository/src/test/java/org/apache/atlas/services/PurgeServiceTest.java
```diff
@@ -0,0 +1,184 @@
...
+        purgeService = Mockito.spy(new PurgeService(atlasGraph, entityStore, typeRegistry));
+    }
+
+    @Test
+    public void testPurgeEntitiesWithNoDeletionCandidates() throws Exception {
+        AtlasIndexQuery mockIndexQuery = mock(AtlasIndexQuery.class);
+        when(atlasGraph.indexQuery(anyString(), anyString())).thenReturn(mockIndexQuery);
+        when(mockIndexQuery.vertices(anyInt(), anyInt())).thenReturn(Collections.<AtlasIndexQuery.Result>emptyList().iterator());
+
+        doReturn(new HashSet<>(Arrays.asList("typeA"))).when(purgeService).getEntityTypes();
```
rkundam: Instead of mocking the entire API response, you can invoke the purge API using AtlasClientV2 and then mock data only where needed in that flow. This ensures we’re validating the actual logic implemented in the Purge service.

With the current approach, purgeEntities() directly returns mock data, which doesn’t allow us to verify whether the method’s flow is executed correctly. The same applies to the other test cases as well.

rkundam (changes requested):

dashboardv3/public/js/views/audit/AdminAuditTableLayoutView.js
```diff
@@ -430,7 +430,7 @@ define(['require',
                 require(['views/audit/AuditTableLayoutView'], function(AuditTableLayoutView) {
                     var obj = {
                             guid: $(e.target).text(),
-                            titleText: (e.target.dataset.operation == "PURGE") ? "Purged Entity Details: " : "Import Details: "
+                            titleText: (e.target.dataset.operation == "PURGE" || e.target.dataset.operation == "AUTO_PURGE") ? "Purged Entity Details: " : "Import Details: "
```
pawarprasad123: Hi,
Looks like for Purge and Auto purge same tille is used "Purge Entity Details", this can be written as 

const titles = {
  PURGE: "Purged Entity Details",
  AUTO_PURGE: "Auto Purge Entity Details"
};

titleText = titles[e.target.dataset.operation] || "Import Details";

dashboardv2/public/js/views/audit/AdminAuditTableLayoutView.js
```diff
@@ -430,7 +430,7 @@ define(['require',
                 require(['views/audit/AuditTableLayoutView'], function(AuditTableLayoutView) {
                     var obj = {
                             guid: $(e.target).text(),
-                            titleText: (e.target.dataset.operation == "PURGE") ? "Purged Entity Details: " : "Import Details: "
+                            titleText: (e.target.dataset.operation == "PURGE" || e.target.dataset.operation == "AUTO_PURGE") ? "Purged Entity Details: " : "Import Details: "
```
pawarprasad123: Hi,
Looks like for Purge and Auto purge same tille is used "Purge Entity Details", this can be written as 

const titles = {
  PURGE: "Purged Entity Details",
  AUTO_PURGE: "Auto Purge Entity Details"
};

titleText = titles[e.target.dataset.operation] || "Import Details";

pawarprasad123 (commented):
UI changes LGTM.

PR #417, opened 2025-07-29
ATLAS-5062: UI:Fix for improving logout mechanism in Atlas UI code base.

pawarprasad123:

[Atlas UI] Fix for improving logout mechanism in Atlas UI code base

Current Behavior :

When a user clicks the logout button from the UI, the system directly navigates to logout.html.
No additional steps are performed after the logout page is hit
New Behavior :

The logout process will be updated to follow a multi-step flow.
Instead of calling logout.html , the new implementation will be calling following APi's and endpoint:
1)checkknoxSSo API
2)lougout API
3)locallogin call
This will ensure a more flexible, maintainable, and robust logout process that aligns with current requirements.

## How was this patch tested?
Manual tested 

(If this patch involves UI changes, please attach a screenshot; otherwise, remove this)
on KNOXafterLogout
afterLogout


mandarambawane (approved):

chaitalicod (approved):

dashboard/src/utils/Utils.ts
```diff
@@ -797,6 +799,66 @@ const globalSearchParams = {
   dslParams: {}
 };
 
+export const handleLogout = async (
```
kevalbhatt, line 802: better to create the logoutUtil?

dashboard/src/views/ErrorPage.tsx
```diff
@@ -16,35 +16,54 @@
...
-          Authentication to this instance of Ranger is managed externally(for
-          example,Apache Knox). You can still open this instance of Ranger from
+          Authentication to this instance of Atlas is managed externally(for
+          example,Apache Knox). You can still open this instance of Atlas from
           the same web browser without re-authentication.To prevent
-          additionalPage not found (404). access to Ranger,
+          additionalPage not found (404). access to Atlas,
           <strong>close all browser windows and exit the browser</strong>.
         </Typography>
       );
```
kevalbhatt, line 45: `useEffect` may not be necessary here since this is derived state. You could compute `errorCode` and `errorInfo` directly from `code` during render to keep things simpler and avoid an extra render.

dashboard/src/views/ErrorPage.tsx
```diff
@@ -74,7 +93,7 @@ export const ErrorPage = (props: { errorCode: string }) => {
           </Stack>
         </Stack>
         <Stack direction="row" spacing={2} className="mt-2">
-          {props.errorCode !== "checkSSOTrue" && (
+          {code !== "checkSSOTrue" && (
```
kevalbhatt, line 96: nit;

```tsx
<CustomButton
  size="small"
  variant="contained"
  color="primary"
  onClick={
    code === "checkSSOTrue"
      ? handleBackWithReload
      : () => navigate("/search")
  }
>
  {code === "checkSSOTrue" ? "Go Back" : "Return to Dashboard"}
</CustomButton>
```

PR #426, opened 2025-08-14
ATLAS-5091: ensure commit/rollback is called on AtlasGraphManagement instances

mneethiraj:

This is a continuatiion of previous patch to address evicted/stale transaction warnings in log. AtlasGraphManagement is now AutoCloseable, which will ensure that commit or rollback is called on all instances, which will avoid transaction leaks and hence evicted/stale transaction warning messages.

## How was this patch tested?
Verified that Atlas startsup successfully, with no errors in loading initial typedef models.

### DIFF
--- a/repository/src/main/java/org/apache/atlas/repository/patches/ReIndexPatch.java
+++ b/repository/src/main/java/org/apache/atlas/repository/patches/ReIndexPatch.java
@@ -26,6 +26,7 @@ import org.apache.atlas.repository.Constants;
 import org.apache.atlas.repository.graphdb.AtlasEdge;
 import org.apache.atlas.repository.graphdb.AtlasElement;
 import org.apache.atlas.repository.graphdb.AtlasGraph;
+import org.apache.atlas.repository.graphdb.AtlasGraphManagement;
 import org.apache.atlas.repository.graphdb.AtlasVertex;
 import org.slf4j.Logger;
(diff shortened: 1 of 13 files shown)


repository/src/main/java/org/apache/atlas/repository/patches/ReIndexPatch.java
```diff
@@ -188,8 +189,10 @@ protected void processItem(AtlasElement item) {
 
         private void attemptCommit() {
             for (String indexName : indexNames) {
-                try {
-                    this.graph.getManagementSystem().reindex(indexName, list);
+                try (AtlasGraphManagement management = this.graph.getManagementSystem()) {
```
pinal-shah, line 192: [RepairIndex.java#L179](https://github.com/apache/atlas/blob/959cb93cece17bcb91f978c41820f92d71a203b1/tools/atlas-index-repair/src/main/java/org/apache/atlas/tools/RepairIndex.java#L179) also needs commit/rrollback.
Although this utility is less preferred to use.

mneethiraj: Reviewed all calls to `JanusGraph.openManagement()`, and updated to ensure that `commit()` is called on the returned instance.

pinal-shah (approved):

PR #470, opened 2025-10-27
ATLAS-5032: Basic search using qualifiedName with startsWith and endsWith operators does not return results for long entity names

aditya-gupta36:

Changes:
Detect attribute values longer than 255 chars and force a graph-query fallback, ensuring Atlas can reliably find entities with very long or tokenized qualifiedNames.

Earlier:
When performing a search using long entity names with attributes like qualifiedName and the StartsWith operator, results are not returned as expected.

Since the qualifiedName attribute is an indexed key, however, Solr's default standard tokenizer has a maximum token length of 255 characters. When entity names exceed this length, the tokenizer fails to parse the value correctly, leading to search failures.

## How was this patch tested?

1. UT Cases
2. UI testing


aditya-gupta36:
Similar Changes merged by another user in [#650](https://github.com/apache/atlas/pull/650)

PR #478, opened 2025-11-25
ATLAS-5156: Import failure due to missing support for handling multip…

sheetalshah1007:
…le relationship types under a single attribute

Atlas handles relationships during import using `EntityGraphMapper`. Earlier, it assumed that an attribute (like `tables`) would have only one relationship type for all its values. Because of this, when different table types (e.g., managed and external tables) were present, it picked a single relationship type for the entire list. This caused issues when the list contained mixed types, leading to incorrect handling during import and updates.

This patch fixes how Atlas handles relationships during import when an attribute contains mixed relationship types.

It introduces a new method, `mapRelationshipAttributeWithMultipleTypes`. Instead of processing all values together, Atlas now:

- Checks whether an attribute can have multiple relationship types
- Splits the values based on each item’s relationship type
- Processes each group separately using the appropriate relationship

In short, rather than treating all values the same, Atlas now correctly handles mixed types. This ensures accurate imports for cases where different table types (such as managed and external tables) coexist, avoiding errors or incorrect relationships.

## How was this patch tested?

A dedicated test class, `EntityGraphMapperMultipleRelationshipTypesTest`, is introduced to:

- Verify mixed table-type collections are correctly split and mapped
- Ensure the correct relationshipDefs are used
- Confirm invalid element types are ignored and do not create relationships

### Summary of Manual Testing:
**Performance**:
Ran large imports (up to 100K tables) on two clusters—with and without the fix—using the same data. Import times are very similar, indicating no performance impact..

**Results**:

1000 tables: ~3m40s → ~3m58s

10000 tables: ~37m → ~38m

50000 tables: ~3h03m → ~3h15m

100000 tables: ~6h44m → ~6h39m

**Multiple relationship types:**
Tested importing a mix of different table types (e.g., managed and external tables). Verified that all types are imported correctly and their relationships are created as expected.

**Partition columns:**
Tested partitioned tables to ensure partition columns and regular columns continue to be handled correctly after the fix.

**Overall correctness:**
Confirmed that all entities and relationships are created correctly, even when different types are mixed together.

**Conclusion**: 
The fix works as expected and does not impact performance. It ensures correct handling of mixed relationship types during import.

### DIFF
--- a/repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
+++ b/repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
@@ -1572,8 +1572,7 @@ public class EntityGraphMapper {
     }
 
     private boolean shouldMapRelationshipAttributeUsingInferredType(Set<String> relationshipTypeNames, Object attrValue) {
-        return CollectionUtils.isEmpty(relationshipTypeNames)
-                || !isAttributeWithMultipleRelationshipTypes(relationshipTypeNames)
+        return !isAttributeWithMultipleRelationshipTypes(relationshipTypeNames)
                 || (!(attrValue instanceof Collection) && !(attrValue instanceof Map));
     }
 
--- a/repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
+++ b/repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
@@ -1532,7 +1532,10 @@ public class EntityGraphMapper {
         Set<String> relationshipTypeNames = entityType.getAttributeRelationshipTypes(attrName);
 
         if (shouldMapRelationshipAttributeUsingInferredType(relationshipTypeNames, attrValue)) {
-            mapRelationshipAttributeUsingInferredType(entityType, attrName, attrValue, vertex, op, context);
+            String relationType = relationshipTypeNames.size() == 1
+                    ? relationshipTypeNames.iterator().next()
+                    : AtlasEntityUtil.getRelationshipType(attrValue);
+            mapRelationshipAttributeUsingInferredType(entityType, attrName, attrValue, relationType, vertex, op, context);
             return;
         }
 
@@ -1576,9 +1579,8 @@ public class EntityGraphMapper {
                 || (!(attrValue instanceof Collection) && !(attrValue instanceof Map));
     }
 
-    private void mapRelationshipAttributeUsingInferredType(AtlasEntityType entityType, String attrName, Object attrValue, AtlasVertex vertex, EntityOperation op, EntityMutationContext context) throws AtlasBaseException {
-        String         relationType = AtlasEntityUtil.getRelationshipType(attrValue);
-        AtlasAttribute attribute    = entityType.getRelationshipAttribute(attrName, relationType);
+    private void mapRelationshipAttributeUsingInferredType(AtlasEntityType entityType, String attrName, Object attrValue, String relationType, AtlasVertex vertex, EntityOperation op, EntityMutationContext context) throws AtlasBaseException {
+        AtlasAttribute attribute = entityType.getRelationshipAttribute(attrName, relationType);
         mapAttribute(attribute, attrValue, vertex, op, context);
     }
 
@@ -1600,13 +1602,10 @@ public class EntityGraphMapper {
     }
 
     private Object createCollectionOfSameType(Object originalValue, List<Object> filteredElements) {
-        if (originalValue instanceof List) {
-            return filteredElements;
-        } else if (originalValue instanceof Set) {
+        if (originalValue instanceof Set) {
             return new HashSet<>(filteredElements);
-        } else {
-            return filteredElements;
         }
+        return filteredElements;
     }
 
     private void mapAttribute(AtlasAttribute attribute, Object attrValue, AtlasVertex vertex, EntityOperation op, EntityMutationContext context) throws AtlasBaseException {
--- a/repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
+++ b/repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
@@ -1507,14 +1507,14 @@ public class EntityGraphMapper {
             if (op.equals(CREATE)) {
                 for (String attrName : entityType.getRelationshipAttributes().keySet()) {
                     Object         attrValue    = entity.getRelationshipAttribute(attrName);
-                    mapRelationshipAttribute(entity, entityType, attrName, attrValue, vertex, op, context);
+                    mapRelationshipAttribute(entityType, attrName, attrValue, vertex, op, context);
                 }
             } else if (op.equals(UPDATE) || op.equals(PARTIAL_UPDATE)) {
                 // relationship attributes mapping
                 for (String attrName : entityType.getRelationshipAttributes().keySet()) {
                     if (entity.hasRelationshipAttribute(attrName)) {
                         Object         attrValue    = entity.getRelationshipAttribute(attrName);
-                        mapRelationshipAttribute(entity, entityType, attrName, attrValue, vertex, op, context);
+                        mapRelationshipAttribute(entityType, attrName, attrValue, vertex, op, context);
                     }
                 }
             }
@@ -1527,11 +1527,11 @@ public class EntityGraphMapper {
         LOG.debug("<== mapRelationshipAttributes({}, {})", op, entity.getTypeName());
     }
 
-    private void mapRelationshipAttribute(AtlasEntity entity, AtlasEntityType entityType, String attrName, Object attrValue, AtlasVertex vertex, EntityOperation op, EntityMutationContext context) throws AtlasBaseException {
-        LOG.debug("==> mapRelationshipAttribute({}, {})", attrName, entity.getTypeName());
+    private void mapRelationshipAttribute(AtlasEntityType entityType, String attrName, Object attrValue, AtlasVertex vertex, EntityOperation op, EntityMutationContext context) throws AtlasBaseException {
+        LOG.debug("==> mapRelationshipAttribute({}, {})", attrName, entityType.getTypeName());
         Set<String> relationshipTypeNames = entityType.getAttributeRelationshipTypes(attrName);
 
-        if (CollectionUtils.isEmpty(relationshipTypeNames)) {
+        if (shouldMapRelationshipAttributeUsingInferredType(relationshipTypeNames, attrValue)) {
             mapRelationshipAttributeUsingInferredType(entityType, attrName, attrValue, vertex, op, context);
             return;
         }
@@ -1560,16 +1560,21 @@ public class EntityGraphMapper {
                 }
             }
         } else if (attrValue instanceof Map) {
-            LOG.warn("mapRelationshipAttribute: attribute {} on {}: Map-valued relationship attribute is not supported", attrName, entity.getTypeName());
-        } else if (relationshipTypeNames.size() == 1) {
-            String         onlyRelationshipType = relationshipTypeNames.iterator().next();
-            AtlasAttribute attribute            = entityType.getRelationshipAttribute(attrName, onlyRelationshipType);
-            mapAttribute(attribute, attrValue, vertex, op, context);
-        } else {
-            mapRelationshipAttributeUsingInferredType(entityType, attrName, attrValue, vertex, op, context);
+            LOG.warn("mapRelationshipAttribute: attribute {} on {}: Map-valued relationship attribute is not supported; use a list or set of references",
+                    attrName, entityType.getTypeName());
         }
 
-        LOG.debug("<== mapRelationshipAttribute({}, {})", attrName, entity.getTypeName());
+        LOG.debug("<== mapRelationshipAttribute({}, {})", attrName, entityType.getTypeName());
+    }
+
+    private boolean isAttributeWithMultipleRelationshipTypes(Set<String> relationshipTypeNames) {
+        return CollectionUtils.isNotEmpty(relationshipTypeNames) && relationshipTypeNames.size() > 1;
+    }
+
+    private boolean shouldMapRelationshipAttributeUsingInferredType(Set<String> relationshipTypeNames, Object attrValue) {
+        return CollectionUtils.isEmpty(relationshipTypeNames)
+                || !isAttributeWithMultipleRelationshipTypes(relationshipTypeNames)
+                || (!(attrValue instanceof Collection) && !(attrValue instanceof Map));
     }
 
     private void mapRelationshipAttributeUsingInferredType(AtlasEntityType entityType, String attrName, Object attrValue, AtlasVertex vertex, EntityOperation op, EntityMutationContext context) throws AtlasBaseException {
@@ -1585,11 +1590,7 @@ public class EntityGraphMapper {
         for (Object element : relatedObjects) {
             String relationshipType = AtlasEntityUtil.getRelationshipType(element);
 
-            if (StringUtils.isEmpty(relationshipType) && relationshipTypeNames != null && relationshipTypeNames.size() == 1) {
-                relationshipType = relationshipTypeNames.iterator().next();
-            }
-
-            if (relationshipTypeNames == null || StringUtils.isEmpty(relationshipType) || !relationshipTypeNames.contains(relationshipType)) {
+            if (StringUtils.isEmpty(relationshipType) || !relationshipTypeNames.contains(relationshipType)) {
                 continue;
             }
 
--- a/repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
+++ b/repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
@@ -1505,22 +1505,19 @@ public class EntityGraphMapper {
             MetricRecorder metric = RequestContext.get().startMetricRecord("mapRelationshipAttributes");
 
             if (op.equals(CREATE)) {
+                // Only map attributes present on the entity; unset names are null and must be skipped (not treated as clears).
                 for (String attrName : entityType.getRelationshipAttributes().keySet()) {
-                    Object         attrValue    = entity.getRelationshipAttribute(attrName);
-                    String         relationType = AtlasEntityUtil.getRelationshipType(attrValue);
-                    AtlasAttribute attribute    = entityType.getRelationshipAttribute(attrName, relationType);
-
-                    mapAttribute(attribute, attrValue, vertex, op, context);
+                    Object attrValue = entity.getRelationshipAttribute(attrName);
+                    if (attrValue != null) {
+                        mapRelationshipAttributeWithMultipleTypes(entity, entityType, attrName, attrValue, vertex, op, context);
+                    }
                 }
             } else if (op.equals(UPDATE) || op.equals(PARTIAL_UPDATE)) {
-                // relationship attributes mapping
+                // relationship attributes mapping — include null when the key is present (explicit clear of that relationship attribute).
                 for (String attrName : entityType.getRelationshipAttributes().keySet()) {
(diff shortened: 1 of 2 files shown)


mandarambawane (approved):

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
```diff
@@ -1498,29 +1498,28 @@ private void mapAttributes(AtlasStruct struct, AtlasStructType structType, Atlas
         LOG.debug("<== mapAttributes({}, {})", op, struct.getTypeName());
     }
 
-    private void mapRelationshipAttributes(AtlasEntity entity, AtlasEntityType entityType, AtlasVertex vertex, EntityOperation op, EntityMutationContext context) throws AtlasBaseException {
+    private void mapRelationshipAttributes(AtlasEntity entity, AtlasEntityType entityType, AtlasVertex vertex, EntityOperation op,
+            EntityMutationContext context) throws AtlasBaseException {
```
pinal-shah: nit

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
```diff
@@ -1498,29 +1498,28 @@ private void mapAttributes(AtlasStruct struct, AtlasStructType structType, Atlas
...
 
             if (op.equals(CREATE)) {
                 for (String attrName : entityType.getRelationshipAttributes().keySet()) {
-                    Object         attrValue    = entity.getRelationshipAttribute(attrName);
-                    String         relationType = AtlasEntityUtil.getRelationshipType(attrValue);
-                    AtlasAttribute attribute    = entityType.getRelationshipAttribute(attrName, relationType);
-
-                    mapAttribute(attribute, attrValue, vertex, op, context);
+                    Object attrValue = entity.getRelationshipAttribute(attrName);
+                    if (attrValue != null) {
```
pinal-shah, line 1511: please review if this check is required

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
```diff
@@ -1498,29 +1498,28 @@ private void mapAttributes(AtlasStruct struct, AtlasStructType structType, Atlas
...
                 // relationship attributes mapping
                 for (String attrName : entityType.getRelationshipAttributes().keySet()) {
                     if (entity.hasRelationshipAttribute(attrName)) {
-                        Object         attrValue    = entity.getRelationshipAttribute(attrName);
-                        String         relationType = AtlasEntityUtil.getRelationshipType(attrValue);
-                        AtlasAttribute attribute    = entityType.getRelationshipAttribute(attrName, relationType);
-
-                        mapAttribute(attribute, attrValue, vertex, op, context);
+                        Object attrValue = entity.getRelationshipAttribute(attrName);
+                        if (attrValue != null) {
```
pinal-shah: please review if this check is required

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
```diff
@@ -1533,6 +1532,98 @@ private void mapRelationshipAttributes(AtlasEntity entity, AtlasEntityType entit
         LOG.debug("<== mapRelationshipAttributes({}, {})", op, entity.getTypeName());
     }
 
+    private void mapRelationshipAttributeWithMultipleTypes(AtlasEntity entity, AtlasEntityType entityType, String attrName, Object attrValue, AtlasVertex vertex, EntityOperation op, EntityMutationContext context) throws AtlasBaseException {
+        LOG.debug("==> mapRelationshipAttributeWithMultipleTypes({}, {})", attrName, entity.getTypeName());
+        Set<String> relationshipTypeNames = entityType.getAttributeRelationshipTypes(attrName);
+
+        if (CollectionUtils.isEmpty(relationshipTypeNames)) {
```
pinal-shah, line 1537: - if relationshipTypeNames is empty, why to map relationship atributes?
- there should be default behaviour incase if 
relationshipTypeNames.size == 1
(!attrValue instanceof Collection && !atrValue instancof Map)

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
```diff
@@ -1533,6 +1532,98 @@ private void mapRelationshipAttributes(AtlasEntity entity, AtlasEntityType entit
...
+            // Fallback to single relationship type processing
+            String         relationType = AtlasEntityUtil.getRelationshipType(attrValue);
+            AtlasAttribute attribute    = entityType.getRelationshipAttribute(attrName, relationType);
+            mapAttribute(attribute, attrValue, vertex, op, context);
+
+            return;
+        }
+
+        if (attrValue instanceof Collection) {
```
pinal-shah, line 1556: Are we making this change considering only icerberg_table, hive_table having same attribut with hive_db? 
If are adding generic code then attrValue can be instance of Map as well

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
```diff
@@ -1533,6 +1532,98 @@ private void mapRelationshipAttributes(AtlasEntity entity, AtlasEntityType entit
...
+            String attrName,
+            Set<String> relationshipTypeNames) {
+        Map<String, List<Object>> elementsByRelationshipType = new HashMap<>();
+
+        // Group related objects by their appropriate relationship type
+        for (Object element : relatedObjects) {
+            // Prefer the explicit relationship type encoded in the element, if any
+            String appropriateRelType = AtlasEntityUtil.getRelationshipType(element);
+
+            if (StringUtils.isEmpty(appropriateRelType)) {
```
pinal-shah: #1593 - #1609
 ```
 if (StringUtils.isEmpty(relationshipType) || !relationshipTypeNames.contains(relationshipType)) {
                           continue;
                    }
```

File: repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java [hunk 7]
```diff
@@ -1533,6 +1532,98 @@ private void mapRelationshipAttributes(AtlasEntity entity, AtlasEntityType entit
...
+            // Process each relationship type with its filtered elements
+            for (Map.Entry<String, List<Object>> entry : elementsByRelationshipType.entrySet()) {
+                String       relationshipTypeName = entry.getKey();
+                List<Object> filteredElements     = entry.getValue();
+
+                AtlasAttribute attribute = entityType.getRelationshipAttribute(attrName, relationshipTypeName);
+
+                if (attribute != null && CollectionUtils.isNotEmpty(filteredElements)) {
+                    // Use the same collection type as the original (List or Set)
+                    Object filteredValue = createCollectionOfSameType(attrValue, filteredElements);
```
pinal-shah, line 1572: how does this matter?

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
```diff
@@ -1533,6 +1532,98 @@ private void mapRelationshipAttributes(AtlasEntity entity, AtlasEntityType entit
...
+
+        if (attrValue instanceof Collection) {
+            Collection<?> relatedObjects = (Collection<?>) attrValue;
+
+            // Group related objects by their appropriate relationship type
+            // e.g., hive_table elements should use hive_table_db relationship, iceberg_table elements should use iceberg_table_db
+            Map<String, List<Object>> elementsByRelationshipType = groupElementsByRelationshipType(
+                    relatedObjects, attrName, relationshipTypeNames);
+
+            // Process each relationship type with its filtered elements
```
pinal-shah: remove unnecessary comments

pinal-shah (changes requested):

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
sheetalshah1007: Done

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
sheetalshah1007: CREATE: We iterate every relationship-attribute name declared on the type. Anything not set on the entity is null. If you removed the` if (attrValue != null) `here, you would call `mapRelationshipAttributeWithMultipleTypes` for every missing attribute and treat “not in payload” like “clear,” which is wrong. So the CREATE null check stays. Comment added for developer's understanding

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
sheetalshah1007: UPDATE / PARTIAL_UPDATE: We only process names where `entity.hasRelationshipAttribute(attrName) `is `true` — the client included that key. The value may be null on purpose to clear that relationship attribute. So we **do not skip** null here; we pass it through so `mapAttribute` can apply defaults or clear, consistent with the old update loop.

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
sheetalshah1007: **Empty relationshipTypeNames:**
`relationshipTypeNames` comes from `getAttributeRelationshipTypes(attrName)`. When it is empty or null, it means we cannot run the new “split by relationship type and map each group” logic — there are no registered relationship-type keys to split on.
That does not mean there is nothing to store. The entity can still carry relationship-attribute values in the request. We still need to write those to the graph.
In that case we use the same behaviour as before this multi-type change: resolve the relationship type from the value` (AtlasEntityUtil.getRelationshipType(attrValue))`, resolve `getRelationshipAttribute(attrName, relationType)`, and call `mapAttribute` once. So “empty `relationshipTypeNames`” means “no multi-type split,” not “skip mapping.” ; we still map using the legacy resolution path.

**Added branch for default behavior:**
If exactly one relationship type and `attrValue` is not a `Collection` and not a `Map`: use that sole type from the registry and `mapAttribute` once (default for a single ref). Comment explains map-shaped single refs skip this and go to else.

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
sheetalshah1007: This change isn’t tied to one combination like “`hive_db + hive_table + iceberg_table.`” It applies whenever one attribute name can point to more than one relationship type. We sort the refs by type and map them in separate steps.

**Maps**:
 Map-valued relationship attribute is not supported yet

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
sheetalshah1007: Simplified loop: 
Resolve relationshipType per element; if empty and only one configured type, set it to that type; then continue if still empty or not in the allowed set.

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
sheetalshah1007: _What createCollectionOfSameType does_:
After splitting by relationship type, `filteredElements` is always built as an `ArrayList`. The client’s original `attrValue` might still be a `List` or a `Set`. This helper makes the value passed into `mapAttribute` match that original kind.

So this line matters because multi-type splitting should not change the collection type of the relationship attribute value—only which refs belong to each relationship type when we call `mapAttribute` for that type.

Added `EntityGraphMapperMultipleRelationshipTypesTest`: contains two flows: same mixed tables refs as `ArrayList`, then as `HashSet`; verifyRelationships each time.

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
sheetalshah1007: Done

sheetalshah1007 (commented):
@pinal-shah All review feedback has been addressed in code, comments, or review replies as appropriate. Please check

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
sheetalshah1007: “Empty” does not mean “no relationship in the request.” It means: the type definition does not list multiple relationship types for that attribute. In that case, we figure out the relationship type from the value.  We only skip mapping if the request doesn’t actually send that relationship attribute.

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
```diff
@@ -1533,6 +1527,88 @@ private void mapRelationshipAttributes(AtlasEntity entity, AtlasEntityType entit
         LOG.debug("<== mapRelationshipAttributes({}, {})", op, entity.getTypeName());
     }
 
+    private void mapRelationshipAttribute(AtlasEntity entity, AtlasEntityType entityType, String attrName, Object attrValue, AtlasVertex vertex, EntityOperation op, EntityMutationContext context) throws AtlasBaseException {
```
pinal-shah: No usage of AtlasEntity entity parameter in mapRelationshipAttribute()
you can use AtlasEntityType entityType

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
```diff
@@ -1533,6 +1527,88 @@ private void mapRelationshipAttributes(AtlasEntity entity, AtlasEntityType entit
...
+                    Object filteredValue = createCollectionOfSameType(attrValue, filteredElements);
+
+                    LOG.debug("Processing relationship type {} for attribute {} with {} elements", relationshipTypeName, attrName, filteredElements.size());
+
+                    mapAttribute(attribute, filteredValue, vertex, op, context);
+                }
+            }
+        } else if (attrValue instanceof Map) {
+            LOG.warn("mapRelationshipAttribute: attribute {} on {}: Map-valued relationship attribute is not supported", attrName, entity.getTypeName());
+        } else if (relationshipTypeNames.size() == 1) {
```
pinal-shah: please review to remove 
```
else if (relationshipTypeNames.size() == 1) {
            String         onlyRelationshipType = relationshipTypeNames.iterator().next();
            AtlasAttribute attribute            = entityType.getRelationshipAttribute(attrName, onlyRelationshipType);
            mapAttribute(attribute, attrValue, vertex, op, context);
        } else {
            mapRelationshipAttributeUsingInferredType(entityType, attrName, attrValue, vertex, op, context);
        }
```

after addressing previous comment

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
```diff
@@ -1533,6 +1527,88 @@ private void mapRelationshipAttributes(AtlasEntity entity, AtlasEntityType entit
...
+    }
+
+    private Map<String, List<Object>> groupElementsByRelationshipType(Collection<?> relatedObjects, String attrName, Set<String> relationshipTypeNames) {
+        Map<String, List<Object>> elementsByRelationshipType = new HashMap<>();
+
+        // Group related objects by their appropriate relationship type
+        for (Object element : relatedObjects) {
+            String relationshipType = AtlasEntityUtil.getRelationshipType(element);
+
+            if (StringUtils.isEmpty(relationshipType) && relationshipTypeNames != null && relationshipTypeNames.size() == 1) {
```
pinal-shah: Please review to remove 1588 to 1590 line after addressing above comments

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
```diff
@@ -1533,6 +1527,88 @@ private void mapRelationshipAttributes(AtlasEntity entity, AtlasEntityType entit
         LOG.debug("<== mapRelationshipAttributes({}, {})", op, entity.getTypeName());
     }
 
+    private void mapRelationshipAttribute(AtlasEntity entity, AtlasEntityType entityType, String attrName, Object attrValue, AtlasVertex vertex, EntityOperation op, EntityMutationContext context) throws AtlasBaseException {
+        LOG.debug("==> mapRelationshipAttribute({}, {})", attrName, entity.getTypeName());
+        Set<String> relationshipTypeNames = entityType.getAttributeRelationshipTypes(attrName);
+
+        if (CollectionUtils.isEmpty(relationshipTypeNames)) {
```
pinal-shah:
```
    if ((CollectionUtils.isNotEmpty(relationshipTypeNames) && relationshipTypeNames.size() == 1) || (!(attrValue instanceof Collection) && !(attrValue instanceof Map)) || CollectionUtils.isEmpty(relationshipTypeNames)) {
        mapRelationshipAttributeUsingInferredType(entityType, attrName, attrValue, vertex, op, context);
        return;
    }
```
Please review if above can be added
Also please add helper method to check condition, something like

isAttributeWithMulipleRelationshipTypes() {}

pinal-shah (changes requested):

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
sheetalshah1007: done

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
sheetalshah1007: Since one method that mixes “how many defs” with “instanceof Collection/Map” would become a long condition that’s harder to name, test, and reuse; i have split the logic into  `isAttributeWithMultipleRelationshipTypes`() and `shouldMapRelationshipAttributeUsingInferredType`() as they are different concerns

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
sheetalshah1007: Done

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
```diff
@@ -1533,6 +1527,89 @@ private void mapRelationshipAttributes(AtlasEntity entity, AtlasEntityType entit
...
+
+        LOG.debug("<== mapRelationshipAttribute({}, {})", attrName, entityType.getTypeName());
+    }
+
+    private boolean isAttributeWithMultipleRelationshipTypes(Set<String> relationshipTypeNames) {
+        return CollectionUtils.isNotEmpty(relationshipTypeNames) && relationshipTypeNames.size() > 1;
+    }
+
+    private boolean shouldMapRelationshipAttributeUsingInferredType(Set<String> relationshipTypeNames, Object attrValue) {
+        return CollectionUtils.isEmpty(relationshipTypeNames)
```
pinal-shah: Please review if this condition is needed 
CollectionUtils.isEmpty(relationshipTypeNames)

pinal-shah (changes requested):

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/EntityGraphMapper.java
sheetalshah1007: Removed it

pinal-shah (approved):

rkundam (approved):

PR #490, opened 2025-12-27
ATLAS-4988: BusinessMetadata with attribute of data type Array takes time to Delete.

UmeshPatil-1:
**ATLAS-4988**
**What changes were proposed in this pull request?**
Background: In Apache Atlas, deleting a Business Metadata (BM) definition requires a pre-check to ensure no existing entities are assigned to those attributes. For attributes with a data type of Array (multi-valued), the deletion process was taking an exceptionally long time (e.g., over 55 minutes for -1,18,300 entities), even when the BM was not assigned to any entity.

**Root Cause:**

1. Array attributes are not Solr-indexed
ARRAY BM attributes are not indexed in Solr.
Solr therefore performs a full collection scan when queried with NOT_EMPTY, causing severe latency.
2. Single-path Solr validation
The previous logic forced all BM attributes (STRING + ARRAY) through Solr.
This was efficient for STRING attributes but pathological for ARRAY attributes.
3. Historical entities ignored
The earlier implementation excluded deleted entities.
This allowed historical BM references to be missed, leading to potential integrity violations.
4. Unnecessary data retrieval
The Solr query requested attribute payloads, increasing I/O cost even though only existence was required.

**Changes Proposed:**

1. Check BM usage in the graph, not in Solr
Before removing a Business Metadata (BM) type, we see if any entity vertex still has the BM property set. That is done with a JanusGraph query on the right vertex property name, instead of using the search index for an “is it used?” check.
2. One graph query instead of two
The “is this BM attribute used?” check used to run two graph queries (direct type match, then supertype match). It now uses one query with an OR: match the entity’s type or any of its super types. We stop after one matching vertex.
3. Honor subtypes when deciding which entity types count
We still expand the configured “applicable entity types” to include subtypes, so a delete does not miss usage on child types. The expansion logic was tightened to go through AtlasEntityType from the type registry.
4. Cache work on the BM type object
AtlasBusinessMetadataType now remembers whether the definition has any non-indexable attribute, and (in a later resolution phase) caches each attribute’s applicable type names including subtypes. Delete uses that cache when present, and falls back to computing from the attribute definition when the cache is empty.
5. force delete for BM only
typedef delete APIs accept force (bulk and delete-by-name). For BM, force=true skips the reference checks and deletes the type anyway (with a warning that orphaned values may stay on entities). The bulk path passes force only into the BM delete loop, other typedef kinds are not supposed to bypass their usual rules because of this flag.
6. Block non-indexable BM unless force is true
If any BM attribute is non-indexable and force is false, delete is rejected with a new error code that tells the user they must use force=true if they accept skipping validation. The rules for treating “indexable” were fixed so null/false on the flag is handled consistently.
7. Plumbing in the typedef store
AtlasDefStore and AtlasAbstractDefStoreV2 gained overloads that carry forceDelete through pre-delete and delete. Non-BM stores keep the old behavior via default methods. deleteByName / deleteByGuid without force now simply call the overload with force=false so there is a single code path.
8. Tests and fixtures
Added and updated tests for BM delete (including mixed indexable / non-indexable and force), REST, client, and graph store.

**Impact**:

Performance Gain: BM typedef delete no longer relies on search for the “is this attribute still used?” step. It uses a small JanusGraph lookup (property present, type or supertype match, stop at one vertex) and reuses cached type lists from AtlasBusinessMetadataType. On large deployments this typically means much shorter and more predictable delete times than the old search-heavy path, especially where attributes were not usefully indexed for that check. 

**API and compatibility**:
Other typedef categories do not get a new “force bypass” through bulk delete; only the BM path uses force as described. Existing clients that omit force keep previous behavior for non-BM deletes.

**Scalability**:
Faster, more predictable BM deletes on big catalogs by using graph existence checks and cached subtype-aware type sets.

**How was this patch tested?**
Maven Build:
Build Successful.

**Manual API tests (curl) BM & Non-BM :**

1. Creation & deletion (no assignment): Created BM with string and array-of-string attributes (indexable), then deleted via DELETE /api/atlas/v2/types/typedefs and DELETE /api/atlas/v2/types/typedef/name/{typeName} with force=false. Expected 204; delete completed.
2. Deletion blocked (with assignment --> indexable BM): Assigned BM to a hive_table (and custom entity types), then tried to remove the BM typedef without force. Deletion was blocked with 409 and TYPE_HAS_REFERENCES.
3. Force delete (Bypass): With BM still assigned, called the same DELETEs with ?force=true. Got 204 and saw the force-delete log path (validation skipped; possible orphaned values on entities).
4. Non-indexable BM: Created BM with isIndexable: false; without force, delete returned 400 / NON_INDEXABLE_BM_DELETE_NOT_ALLOWED; with force=true, delete returned 204.
5. Non-BM typedefs (classifications and other categories): Used bulk DELETE /api/atlas/v2/types/typedefs for non-BM typedefs with force=false vs force=true. No instances --> 204 in both cases; instances still present --> 409 in both cases
 --> force does not bypass reference checks outside BM.
 
 
**API Usage Note: force on typedef delete (ATLAS-4988)**
**Supported Endpoints**
**Bulk delete**
DELETE /api/atlas/v2/types/typedefs?force=true|false
**Delete by name**
DELETE /api/atlas/v2/types/typedef/name/{typeName}?force=true|false
force is optional and defaults to false.

**Important: force only affects Business Metadata typedefs**
The force parameter is meaningful only for deleting Business Metadata (businessMetadataDefs) typedefs.
When deleting a Business Metadata type:
force=false (default)
Performs the standard pre-delete reference validation.
If the BM is still assigned to entities, deletion fails with reference-related errors.
force=true
Skips the pre-delete reference validation and deletes the Business Metadata typedef even if references still exist.
This may leave orphaned BM values on entities.
**Non-Business Metadata typedefs**
For all other typedef categories (enum, struct, classification, entity, relationship, etc.), the force parameter does not bypass existing validation or reference checks.
Behavior for these typedefs remains unchanged from previous releases:
Types without references can still be deleted successfully.
Types with existing references or instances continue to fail with errors such as:
409 / TYPE_HAS_REFERENCES
This behavior is consistent regardless of whether force=true or force=false is supplied.

### DIFF
--- a/client/client-v2/src/main/java/org/apache/atlas/AtlasClientV2.java
+++ b/client/client-v2/src/main/java/org/apache/atlas/AtlasClientV2.java
@@ -353,7 +353,19 @@ public class AtlasClientV2 extends AtlasBaseClient {
     }
 
     public void deleteTypeByName(String typeName) throws AtlasServiceException {
-        callAPI(API_V2.DELETE_TYPE_DEF_BY_NAME, (Class) null, null, typeName);
+        callAPI(API_V2.DELETE_TYPE_DEF_BY_NAME, (Class<?>) null, null, typeName);
+    }
+
+    public void deleteTypeByName(String typeName, boolean forceDelete) throws AtlasServiceException {
+        if (forceDelete) {
+
+            MultivaluedMap<String, String> queryParams = new MultivaluedMapImpl();
+            queryParams.add("force", "true");
+            callAPI(API_V2.DELETE_TYPE_DEF_BY_NAME, (Class<?>) null, queryParams, typeName);
+        } else {
+
+            callAPI(API_V2.DELETE_TYPE_DEF_BY_NAME, (Class<?>) null, null, typeName);
+        }
     }
 
     // Entity APIs
--- a/client/client-v2/src/main/java/org/apache/atlas/AtlasClientV2.java
+++ b/client/client-v2/src/main/java/org/apache/atlas/AtlasClientV2.java
@@ -357,15 +357,14 @@ public class AtlasClientV2 extends AtlasBaseClient {
     }
 
     public void deleteTypeByName(String typeName, boolean forceDelete) throws AtlasServiceException {
+        MultivaluedMap<String, String> queryParams = null;
+
         if (forceDelete) {
-
-            MultivaluedMap<String, String> queryParams = new MultivaluedMapImpl();
+            queryParams = new MultivaluedMapImpl();
             queryParams.add("force", "true");
-            callAPI(API_V2.DELETE_TYPE_DEF_BY_NAME, (Class<?>) null, queryParams, typeName);
-        } else {
-
-            callAPI(API_V2.DELETE_TYPE_DEF_BY_NAME, (Class<?>) null, null, typeName);
         }
+
+        callAPI(API_V2.DELETE_TYPE_DEF_BY_NAME, (Class<?>) null, queryParams, typeName);
     }
 
     // Entity APIs
--- a/client/client-v2/src/main/java/org/apache/atlas/AtlasClientV2.java
+++ b/client/client-v2/src/main/java/org/apache/atlas/AtlasClientV2.java
@@ -349,7 +349,18 @@ public class AtlasClientV2 extends AtlasBaseClient {
      * @param typesDef A composite object that captures all types to be deleted
      */
     public void deleteAtlasTypeDefs(AtlasTypesDef typesDef) throws AtlasServiceException {
-        callAPI(API_V2.DELETE_TYPE_DEFS, (Class<?>) null, AtlasType.toJson(typesDef));
+        deleteAtlasTypeDefs(typesDef, false);
+    }
+
+    public void deleteAtlasTypeDefs(AtlasTypesDef typesDef, boolean forceDelete) throws AtlasServiceException {
+        MultivaluedMap<String, String> queryParams = null;
+
+        if (forceDelete) {
+            queryParams = new MultivaluedMapImpl();
+            queryParams.add("force", "true");
+        }
+
+        callAPI(API_V2.DELETE_TYPE_DEFS, (Class<?>) null, AtlasType.toJson(typesDef), queryParams);
     }
 
     public void deleteTypeByName(String typeName) throws AtlasServiceException {
--- a/intg/src/main/java/org/apache/atlas/AtlasErrorCode.java
+++ b/intg/src/main/java/org/apache/atlas/AtlasErrorCode.java
@@ -261,7 +261,7 @@ public enum AtlasErrorCode {
     ABORT_IMPORT_FAILED(500, "ATLAS-500-00-022", "Failed to abort import with id {0}"),
     IMPORT_QUEUEING_FAILED(500, "ATLAS-500-00-023", "Failed to add import with id {0} to request queue, please try again later"),
 
-    NON_INDEXABLE_BM_DELETE_NOT_ALLOWED(400, "ATLAS-400-00-106", "Deletion not allowed for non-indexable Business Metadata ''{0}'' without force-delete. Please use the force-delete parameter to remove.");
+    NON_INDEXABLE_BM_DELETE_NOT_ALLOWED(400, "ATLAS-400-00-106", "Deletion not allowed for non-indexable Business Metadata ''{0}'' without force=true. Non-indexable attributes cannot be validated efficiently for references; use force=true to skip validation and delete (warning: orphaned references may remain).");
 
     private static final Logger LOG = LoggerFactory.getLogger(AtlasErrorCode.class);
     private final        String errorCode;
--- a/intg/src/main/java/org/apache/atlas/AtlasErrorCode.java
+++ b/intg/src/main/java/org/apache/atlas/AtlasErrorCode.java
@@ -183,6 +183,7 @@ public enum AtlasErrorCode {
     BLANK_NAME_ATTRIBUTE(400, "ATLAS-400-00-104", "Name Attribute can't be empty!"),
     BLANK_VALUE_ATTRIBUTE(400, "ATLAS-400-00-105", "Value Attribute can't be empty!"),
     INVALID_RELATIONSHIP_LABEL(400, "ATLAS-400-00-106", "Invalid relationship label {0}. The referenced entity type {1} could not be resolved from the type registry."),
+    NON_INDEXABLE_BM_DELETE_NOT_ALLOWED(400, "ATLAS-400-00-107", "Deletion not allowed for non-indexable Business Metadata ''{0}'' without force=true. Non-indexable attributes cannot be validated efficiently for references; use force=true to skip validation and delete (warning: orphaned references may remain)."),
 
     UNAUTHORIZED_ACCESS(403, "ATLAS-403-00-001", "{0} is not authorized to perform {1}"),
 
@@ -259,9 +260,7 @@ public enum AtlasErrorCode {
     IMPORT_REGISTRATION_FAILED(500, "ATLAS-500-00-020", "Failed to register import request"),
     IMPORT_FAILED(500, "ATLAS-500-00-021", "Import with id {0} failed"),
     ABORT_IMPORT_FAILED(500, "ATLAS-500-00-022", "Failed to abort import with id {0}"),
-    IMPORT_QUEUEING_FAILED(500, "ATLAS-500-00-023", "Failed to add import with id {0} to request queue, please try again later"),
-
-    NON_INDEXABLE_BM_DELETE_NOT_ALLOWED(400, "ATLAS-400-00-106", "Deletion not allowed for non-indexable Business Metadata ''{0}'' without force=true. Non-indexable attributes cannot be validated efficiently for references; use force=true to skip validation and delete (warning: orphaned references may remain).");
+    IMPORT_QUEUEING_FAILED(500, "ATLAS-500-00-023", "Failed to add import with id {0} to request queue, please try again later");
 
     private static final Logger LOG = LoggerFactory.getLogger(AtlasErrorCode.class);
     private final        String errorCode;
--- a/intg/src/main/java/org/apache/atlas/AtlasErrorCode.java
+++ b/intg/src/main/java/org/apache/atlas/AtlasErrorCode.java
@@ -259,7 +259,9 @@ public enum AtlasErrorCode {
     IMPORT_REGISTRATION_FAILED(500, "ATLAS-500-00-020", "Failed to register import request"),
     IMPORT_FAILED(500, "ATLAS-500-00-021", "Import with id {0} failed"),
     ABORT_IMPORT_FAILED(500, "ATLAS-500-00-022", "Failed to abort import with id {0}"),
-    IMPORT_QUEUEING_FAILED(500, "ATLAS-500-00-023", "Failed to add import with id {0} to request queue, please try again later");
+    IMPORT_QUEUEING_FAILED(500, "ATLAS-500-00-023", "Failed to add import with id {0} to request queue, please try again later"),
+
+    NON_INDEXABLE_BM_DELETE_NOT_ALLOWED(400, "ATLAS-400-00-106", "Deletion not allowed for non-indexable Business Metadata ''{0}'' without force-delete. Please use the force-delete parameter to remove.");
 
     private static final Logger LOG = LoggerFactory.getLogger(AtlasErrorCode.class);
     private final        String errorCode;
--- a/intg/src/main/java/org/apache/atlas/type/AtlasBusinessMetadataType.java
+++ b/intg/src/main/java/org/apache/atlas/type/AtlasBusinessMetadataType.java
@@ -44,10 +44,16 @@ public class AtlasBusinessMetadataType extends AtlasStructType {
 
     private final AtlasBusinessMetadataDef businessMetadataDef;
 
+    private boolean hasNonIndexableAttributes;
+
     public AtlasBusinessMetadataType(AtlasBusinessMetadataDef businessMetadataDef) {
         super(businessMetadataDef);
 
         this.businessMetadataDef = businessMetadataDef;
+    }
+
+    public boolean hasNonIndexableAttributes() {
+        return hasNonIndexableAttributes;
     }
 
     @Override
@@ -70,6 +76,8 @@ public class AtlasBusinessMetadataType extends AtlasStructType {
         super.resolveReferences(typeRegistry);
 
         Map<String, AtlasBusinessAttribute> a = new HashMap<>();
+
+        hasNonIndexableAttributes = false;
 
         for (AtlasAttribute attribute : super.allAttributes.values()) {
             AtlasAttributeDef attributeDef = attribute.getAttributeDef();
@@ -115,10 +123,16 @@ public class AtlasBusinessMetadataType extends AtlasStructType {
                 bmAttribute = new AtlasBusinessAttribute(attribute, entityTypes);
             }
 
+            if (!hasNonIndexableAttributes && !Boolean.TRUE.equals(attributeDef.getIsIndexable())) {
+                hasNonIndexableAttributes = true;
+            }
+
             a.put(attrName, bmAttribute);
         }
 
         super.allAttributes = Collections.unmodifiableMap(a);
+        LOG.debug("CACHE-POPULATE [Phase1] BM='{}' hasNonIndexableAttributes={} totalAttributes={}",
+                getTypeName(), hasNonIndexableAttributes, a.size());
     }
 
     @Override
@@ -137,6 +151,29 @@ public class AtlasBusinessMetadataType extends AtlasStructType {
         }
     }
 
+    @Override
+    void resolveReferencesPhase3(AtlasTypeRegistry typeRegistry) throws AtlasBaseException {
+        super.resolveReferencesPhase3(typeRegistry);
+
+        for (AtlasAttribute attribute : super.allAttributes.values()) {
+            AtlasBusinessAttribute bmAttribute = (AtlasBusinessAttribute) attribute;
+            Set<AtlasEntityType>   entityTypes = bmAttribute.getApplicableEntityTypes();
+
+            if (CollectionUtils.isNotEmpty(entityTypes)) {
+                Set<String> expandedTypeNames = new HashSet<>();
+
+                for (AtlasEntityType entityType : entityTypes) {
+                    expandedTypeNames.add(entityType.getTypeName());
+                    expandedTypeNames.addAll(entityType.getAllSubTypes());
+                }
+
+                bmAttribute.applicableEntityTypeNamesWithSubTypes = Collections.unmodifiableSet(expandedTypeNames);
+                LOG.debug("CACHE-POPULATE [Phase3] BM='{}' attr='{}' expandedTypes={}",
+                        getTypeName(), bmAttribute.getName(), expandedTypeNames);
+            }
+        }
+    }
+
     public AtlasBusinessMetadataDef getBusinessMetadataDef() {
         return businessMetadataDef;
     }
@@ -145,6 +182,8 @@ public class AtlasBusinessMetadataType extends AtlasStructType {
         private final Set<AtlasEntityType> applicableEntityTypes;
         private final int                  maxStringLength;
         private final String               validPattern;
+
+        private Set<String> applicableEntityTypeNamesWithSubTypes = Collections.emptySet();
 
         public AtlasBusinessAttribute(AtlasAttribute attribute, Set<AtlasEntityType> applicableEntityTypes) {
             super(attribute);
@@ -169,6 +208,10 @@ public class AtlasBusinessMetadataType extends AtlasStructType {
 
         public Set<AtlasEntityType> getApplicableEntityTypes() {
             return applicableEntityTypes;
+        }
+
+        public Set<String> getApplicableEntityTypeNamesWithSubTypes() {
+            return applicableEntityTypeNamesWithSubTypes;
         }
 
         public String getValidPattern() {
--- a/intg/src/main/java/org/apache/atlas/type/AtlasBusinessMetadataType.java
+++ b/intg/src/main/java/org/apache/atlas/type/AtlasBusinessMetadataType.java
@@ -123,7 +123,7 @@ public class AtlasBusinessMetadataType extends AtlasStructType {
                 bmAttribute = new AtlasBusinessAttribute(attribute, entityTypes);
             }
 
-            if (!hasNonIndexableAttributes && !Boolean.TRUE.equals(attributeDef.getIsIndexable())) {
+            if (!hasNonIndexableAttributes && !attributeDef.getIsIndexable()) {
                 hasNonIndexableAttributes = true;
             }
 
--- a/repository/src/main/java/org/apache/atlas/repository/store/graph/AtlasDefStore.java
+++ b/repository/src/main/java/org/apache/atlas/repository/store/graph/AtlasDefStore.java
@@ -45,9 +45,25 @@ public interface AtlasDefStore<T extends AtlasBaseTypeDef> {
 
     AtlasVertex preDeleteByName(String name) throws AtlasBaseException;
 
-    void deleteByName(String name, AtlasVertex preDeleteResult) throws AtlasBaseException;
-
     AtlasVertex preDeleteByGuid(String guid) throws AtlasBaseException;
 
+    void deleteByName(String name, AtlasVertex preDeleteResult) throws AtlasBaseException;
+
     void deleteByGuid(String guid, AtlasVertex preDeleteResult) throws AtlasBaseException;
+
+    default AtlasVertex preDeleteByName(String name, boolean forceDelete) throws AtlasBaseException {
+        return preDeleteByName(name);
+    }
+
+    default void deleteByName(String name, AtlasVertex preDeleteResult, boolean forceDelete) throws AtlasBaseException {
+        deleteByName(name, preDeleteResult);
+    }
+
+    default AtlasVertex preDeleteByGuid(String guid, boolean forceDelete) throws AtlasBaseException {
+        return preDeleteByGuid(guid);
+    }
+
+    default void deleteByGuid(String guid, AtlasVertex preDeleteResult, boolean forceDelete) throws AtlasBaseException {
+        deleteByGuid(guid, preDeleteResult);
+    }
 }
--- a/repository/src/main/java/org/apache/atlas/repository/store/graph/AtlasDefStore.java
+++ b/repository/src/main/java/org/apache/atlas/repository/store/graph/AtlasDefStore.java
@@ -20,6 +20,8 @@ package org.apache.atlas.repository.store.graph;
 import org.apache.atlas.exception.AtlasBaseException;
 import org.apache.atlas.model.typedef.AtlasBaseTypeDef;
 import org.apache.atlas.repository.graphdb.AtlasVertex;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.List;
 
@@ -27,6 +29,8 @@ import java.util.List;
  * Interface for graph persistence store for AtlasTypeDef
  */
 public interface AtlasDefStore<T extends AtlasBaseTypeDef> {
+    Logger LOG = LoggerFactory.getLogger(AtlasDefStore.class);
+
     AtlasVertex preCreate(T typeDef) throws AtlasBaseException;
 
     T create(T typeDef, AtlasVertex preCreateResult) throws AtlasBaseException;
@@ -52,6 +56,10 @@ public interface AtlasDefStore<T extends AtlasBaseTypeDef> {
     void deleteByGuid(String guid, AtlasVertex preDeleteResult) throws AtlasBaseException;
 
     default AtlasVertex preDeleteByName(String name, boolean forceDelete) throws AtlasBaseException {
+        if (forceDelete) {
+            LOG.debug("Force-delete flag ignored in {}.preDeleteByName() for type '{}'; feature is implemented only for BusinessMetadata.", getClass().getSimpleName(), name);
+        }
+
         return preDeleteByName(name);
     }
 
@@ -60,6 +68,10 @@ public interface AtlasDefStore<T extends AtlasBaseTypeDef> {
     }
 
     default AtlasVertex preDeleteByGuid(String guid, boolean forceDelete) throws AtlasBaseException {
+        if (forceDelete) {
+            LOG.debug("Force-delete flag ignored in {}.preDeleteByGuid() for guid '{}'; feature is implemented only for BusinessMetadata.", getClass().getSimpleName(), guid);
+        }
+
         return preDeleteByGuid(guid);
     }
 
--- a/repository/src/main/java/org/apache/atlas/repository/store/graph/AtlasDefStore.java
+++ b/repository/src/main/java/org/apache/atlas/repository/store/graph/AtlasDefStore.java
@@ -20,8 +20,6 @@ package org.apache.atlas.repository.store.graph;
 import org.apache.atlas.exception.AtlasBaseException;
 import org.apache.atlas.model.typedef.AtlasBaseTypeDef;
 import org.apache.atlas.repository.graphdb.AtlasVertex;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import java.util.List;
 
@@ -29,8 +27,6 @@ import java.util.List;
  * Interface for graph persistence store for AtlasTypeDef
  */
 public interface AtlasDefStore<T extends AtlasBaseTypeDef> {
-    Logger LOG = LoggerFactory.getLogger(AtlasDefStore.class);
-
     AtlasVertex preCreate(T typeDef) throws AtlasBaseException;
 
     T create(T typeDef, AtlasVertex preCreateResult) throws AtlasBaseException;
@@ -56,10 +52,6 @@ public interface AtlasDefStore<T extends AtlasBaseTypeDef> {
     void deleteByGuid(String guid, AtlasVertex preDeleteResult) throws AtlasBaseException;
 
     default AtlasVertex preDeleteByName(String name, boolean forceDelete) throws AtlasBaseException {
-        if (forceDelete) {
-            LOG.debug("Force-delete flag ignored in {}.preDeleteByName() for type '{}'; feature is implemented only for BusinessMetadata.", getClass().getSimpleName(), name);
-        }
-
         return preDeleteByName(name);
     }
 
@@ -68,10 +60,6 @@ public interface AtlasDefStore<T extends AtlasBaseTypeDef> {
     }
 
     default AtlasVertex preDeleteByGuid(String guid, boolean forceDelete) throws AtlasBaseException {
-        if (forceDelete) {
-            LOG.debug("Force-delete flag ignored in {}.preDeleteByGuid() for guid '{}'; feature is implemented only for BusinessMetadata.", getClass().getSimpleName(), guid);
-        }
-
         return preDeleteByGuid(guid);
     }
 
--- a/repository/src/main/java/org/apache/atlas/repository/store/graph/AtlasTypeDefGraphStore.java
+++ b/repository/src/main/java/org/apache/atlas/repository/store/graph/AtlasTypeDefGraphStore.java
@@ -539,14 +539,20 @@ public abstract class AtlasTypeDefGraphStore implements AtlasTypeDefStore {
     @Override
     @GraphTransaction
     public void deleteTypesDef(AtlasTypesDef typesDef) throws AtlasBaseException {
+        deleteTypesDef(typesDef, false);
+    }
+
+    @GraphTransaction
+    public void deleteTypesDef(AtlasTypesDef typesDef, boolean forceDelete) throws AtlasBaseException {
         if (LOG.isDebugEnabled()) {
-            LOG.debug("==> AtlasTypeDefGraphStore.deleteTypesDef(enums={}, structs={}, classfications={}, entities={}, relationships={}, businessMetadataDefs={})",
+            LOG.debug("==> AtlasTypeDefGraphStore.deleteTypesDef(enums={}, structs={}, classfications={}, entities={}, relationships={}, businessMetadataDefs={}, forceDelete={})",
                     CollectionUtils.size(typesDef.getEnumDefs()),
                     CollectionUtils.size(typesDef.getStructDefs()),
                     CollectionUtils.size(typesDef.getClassificationDefs()),
                     CollectionUtils.size(typesDef.getEntityDefs()),
                     CollectionUtils.size(typesDef.getRelationshipDefs()),
-                    CollectionUtils.size(typesDef.getBusinessMetadataDefs()));
+                    CollectionUtils.size(typesDef.getBusinessMetadataDefs()),
+                    forceDelete);
         }
 
         AtlasTransientTypeRegistry ttr = lockTypeRegistryAndReleasePostCommit();
@@ -678,9 +684,9 @@ public abstract class AtlasTypeDefGraphStore implements AtlasTypeDefStore {
         if (CollectionUtils.isNotEmpty(typesDef.getBusinessMetadataDefs())) {
             for (AtlasBusinessMetadataDef businessMetadataDef : typesDef.getBusinessMetadataDefs()) {
                 if (StringUtils.isNotBlank(businessMetadataDef.getGuid())) {
-                    businessMetadataDefStore.deleteByGuid(businessMetadataDef.getGuid(), null);
+                    businessMetadataDefStore.deleteByGuid(businessMetadataDef.getGuid(), null, forceDelete);
                 } else {
-                    businessMetadataDefStore.deleteByName(businessMetadataDef.getName(), null);
+                    businessMetadataDefStore.deleteByName(businessMetadataDef.getName(), null, forceDelete);
                 }
             }
         }
@@ -777,7 +783,7 @@ public abstract class AtlasTypeDefGraphStore implements AtlasTypeDefStore {
 
     @Override
     @GraphTransaction
-    public void deleteTypeByName(String typeName) throws AtlasBaseException {
+    public void deleteTypeByName(String typeName, boolean forceDelete) throws AtlasBaseException {
         AtlasType atlasType = typeRegistry.getType(typeName);
 
         if (atlasType == null) {
@@ -801,7 +807,7 @@ public abstract class AtlasTypeDefGraphStore implements AtlasTypeDefStore {
             typesDef.setStructDefs(Collections.singletonList((AtlasStructDef) baseTypeDef));
         }
 
-        deleteTypesDef(typesDef);
+        deleteTypesDef(typesDef, forceDelete);
     }
 
     @Override
--- a/repository/src/main/java/org/apache/atlas/repository/store/graph/AtlasTypeDefGraphStore.java
+++ b/repository/src/main/java/org/apache/atlas/repository/store/graph/AtlasTypeDefGraphStore.java
@@ -544,16 +544,14 @@ public abstract class AtlasTypeDefGraphStore implements AtlasTypeDefStore {
 
     @GraphTransaction
     public void deleteTypesDef(AtlasTypesDef typesDef, boolean forceDelete) throws AtlasBaseException {
-        if (LOG.isDebugEnabled()) {
-            LOG.debug("==> AtlasTypeDefGraphStore.deleteTypesDef(enums={}, structs={}, classfications={}, entities={}, relationships={}, businessMetadataDefs={}, forceDelete={})",
-                    CollectionUtils.size(typesDef.getEnumDefs()),
-                    CollectionUtils.size(typesDef.getStructDefs()),
-                    CollectionUtils.size(typesDef.getClassificationDefs()),
-                    CollectionUtils.size(typesDef.getEntityDefs()),
-                    CollectionUtils.size(typesDef.getRelationshipDefs()),
-                    CollectionUtils.size(typesDef.getBusinessMetadataDefs()),
-                    forceDelete);
-        }
+        LOG.debug("==> AtlasTypeDefGraphStore.deleteTypesDef(enums={}, structs={}, classfications={}, entities={}, relationships={}, businessMetadataDefs={}, forceDelete={})",
+                CollectionUtils.size(typesDef.getEnumDefs()),
+                CollectionUtils.size(typesDef.getStructDefs()),
+                CollectionUtils.size(typesDef.getClassificationDefs()),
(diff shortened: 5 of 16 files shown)


repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasBusinessMetadataDefStoreV2.java
```diff
@@ -363,44 +368,71 @@ private AtlasBusinessMetadataDef toBusinessMetadataDef(AtlasVertex vertex) throw
...
-            List<AtlasStructDef.AtlasAttributeDef> attributeDefs = businessMetadataDef.getAttributeDefs();
+        if (businessMetadataDef == null || CollectionUtils.isEmpty(businessMetadataDef.getAttributeDefs())) {
+            return;
+        }
 
-            for (AtlasStructDef.AtlasAttributeDef attributeDef : attributeDefs) {
-                String      qualifiedName      = AtlasStructType.AtlasAttribute.getQualifiedAttributeName(businessMetadataDef, attributeDef.getName());
-                String      vertexPropertyName = AtlasStructType.AtlasAttribute.generateVertexPropertyName(businessMetadataDef, attributeDef, qualifiedName);
-                Set<String> applicableTypes    = AtlasJson.fromJson(attributeDef.getOption(AtlasBusinessMetadataDef.ATTR_OPTION_APPLICABLE_ENTITY_TYPES), Set.class);
+        for (AtlasStructDef.AtlasAttributeDef attributeDef : businessMetadataDef.getAttributeDefs()) {
```
sheetalshah1007, line 415: Extract the attribute iteration logic in `checkBusinessMetadataRef` to a private method (e.g., `validateAttributeReferences`) for better readability.

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasBusinessMetadataDefStoreV2.java
```diff
@@ -363,44 +368,71 @@ private AtlasBusinessMetadataDef toBusinessMetadataDef(AtlasVertex vertex) throw
...
-        searchParameters.setAttributes(Collections.singleton(attrName));
-        searchParameters.setOffset(0);
+        searchParameters.setAttributes(Collections.emptySet());
         searchParameters.setLimit(1);
 
         AtlasSearchResult atlasSearchResult = entityDiscoveryService.searchWithParameters(searchParameters);
+        return atlasSearchResult != null && CollectionUtils.isNotEmpty(atlasSearchResult.getEntities());
+    }
+
+    private boolean isBusinessAttributePresentInGraph(String vertexPropertyName, Set<String> applicableTypes) {
```
sheetalshah1007: **1. Replace manual iteration with Graph Query for better performance:**
The method isBusinessAttributePresentInGraph manually iterates over all vertices of each applicable type using 	`graph.getVertices(Constants.ENTITY_TYPE_PROPERTY_KEY, typeName), `then checks each vertex for the property and state. This can be slow for large graphs, as it loads and inspects every vertex in memory. This needs to be improved

**2. Add error handling and logging:**
No exception handling in the graph operations. 
If the graph is unavailable or queries fail, it could lead to unhandled exceptions. Wrap the query in a try-catch and log warnings/errors. 
Also, add optional performance logging to monitor query times in production.

**3. Add JUnit tests for the new method covering:**
-Non-indexable attributes with/without assignments.
-Inactive entities (should not count).

sheetalshah1007 (changes requested):

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasBusinessMetadataDefStoreV2.java
UmeshPatil-1: Done

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasBusinessMetadataDefStoreV2.java
```diff
@@ -363,44 +369,91 @@ private AtlasBusinessMetadataDef toBusinessMetadataDef(AtlasVertex vertex) throw
...
+        if (graph == null || CollectionUtils.isEmpty(applicableTypes)) {
+            return false;
+        }
+
+        try {
+            for (String typeName : applicableTypes) {
 
-        return CollectionUtils.isNotEmpty(atlasSearchResult.getEntities());
+                Iterable<AtlasVertex> vertices = graph.query()
+                        .has(Constants.ENTITY_TYPE_PROPERTY_KEY, typeName)
```
mneethiraj: Is it necessary to add `ENTITY_TYPE_PROPERTY_KEY` filter here? If yes, note that all subTypes of types in`applicableTypes` should be checked as well; consider a business attribute associated with a type like `DataSet`.

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasBusinessMetadataDefStoreV2.java
```diff
@@ -363,44 +369,91 @@ private AtlasBusinessMetadataDef toBusinessMetadataDef(AtlasVertex vertex) throw
...
+                        .has(vertexPropertyName, AtlasGraphQuery.ComparisionOperator.NOT_EQUAL, (Object) null)
+                        .has(Constants.STATE_PROPERTY_KEY, AtlasEntity.Status.ACTIVE.name())
```
mneethiraj: Is it necessary to check only for ACTIVE entities? The business metadata can't be deleted even when a deleted entity has reference to it, right?

UmeshPatil-1: Thank's for pointing these out, To ensure full data integrity, we should block the deletion of Business Metadata if any entity—Active or Deleted—still holds a reference to it.

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasBusinessMetadataDefStoreV2.java
```diff
@@ -363,44 +371,111 @@ private AtlasBusinessMetadataDef toBusinessMetadataDef(AtlasVertex vertex) throw
     private void checkBusinessMetadataRef(String typeName) throws AtlasBaseException {
```
pinal-shah: Consider introducing option of forceDelete

And check references only if !forceDelete and isIndexable = true

client/client-v2/src/main/java/org/apache/atlas/AtlasClientV2.java
```diff
@@ -353,7 +353,19 @@ public void deleteAtlasTypeDefs(AtlasTypesDef typesDef) throws AtlasServiceExcep
     }
 
     public void deleteTypeByName(String typeName) throws AtlasServiceException {
-        callAPI(API_V2.DELETE_TYPE_DEF_BY_NAME, (Class) null, null, typeName);
+        callAPI(API_V2.DELETE_TYPE_DEF_BY_NAME, (Class<?>) null, null, typeName);
```
sheetalshah1007, line 367: The if-else with duplicate API calls can be avoided; consider combining the if-else logic since both overloaded methods call the same API - just with different query params.

repository/src/main/java/org/apache/atlas/repository/store/graph/AtlasDefStore.java
```diff
@@ -45,9 +45,25 @@ public interface AtlasDefStore<T extends AtlasBaseTypeDef> {
...
+    }
+
+    default AtlasVertex preDeleteByGuid(String guid, boolean forceDelete) throws AtlasBaseException {
+        return preDeleteByGuid(guid);
+    }
+
+    default void deleteByGuid(String guid, AtlasVertex preDeleteResult, boolean forceDelete) throws AtlasBaseException {
+        deleteByGuid(guid, preDeleteResult);
+    }
 }
```
sheetalshah1007, line 69: Good use of default methods for backward compatibility.

SUGGESTION: Add debug logging when force-delete flag is ignored. Force-delete only applies to BusinessMetadata (fixes slow array deletion); the other 5 type stores (Enum, Struct, Classification, Entity, Relationship) don't need it, so these default methods just ignore the flag - which is fine.
Adding a log like "Force-delete flag ignored for {typeName} - BusinessMetadata-only feature" clarifies the design and helps debug deletion issues without any breaking changes.

intg/src/main/java/org/apache/atlas/AtlasErrorCode.java
```diff
@@ -257,7 +257,9 @@ public enum AtlasErrorCode {
     IMPORT_REGISTRATION_FAILED(500, "ATLAS-500-00-020", "Failed to register import request"),
     IMPORT_FAILED(500, "ATLAS-500-00-021", "Import with id {0} failed"),
     ABORT_IMPORT_FAILED(500, "ATLAS-500-00-022", "Failed to abort import with id {0}"),
-    IMPORT_QUEUEING_FAILED(500, "ATLAS-500-00-023", "Failed to add import with id {0} to request queue, please try again later");
+    IMPORT_QUEUEING_FAILED(500, "ATLAS-500-00-023", "Failed to add import with id {0} to request queue, please try again later"),
+
+    NON_INDEXABLE_BM_DELETE_NOT_ALLOWED(400, "ATLAS-400-00-106", "Deletion not allowed for non-indexable Business Metadata ''{0}'' without force-delete. Please use the force-delete parameter to remove.");
```
sheetalshah1007: the message could be more user-friendly by explaining WHAT force-delete does.
Suggested message:
"Deletion not allowed for non-indexable Business Metadata ''{0}'' without force parameter.
Non-indexable attributes cannot be efficiently validated for references. Use force=true to
skip validation and delete (Warning: may leave orphaned references)."

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasBusinessMetadataDefStoreV2.java
```diff
@@ -291,15 +319,35 @@ public AtlasVertex preDeleteByGuid(String guid) throws AtlasBaseException {
             throw new AtlasBaseException(AtlasErrorCode.TYPE_GUID_NOT_FOUND, guid);
         }
 
-        if (existingDef != null) {
+        if (existingDef != null && !forceDelete) {
+            boolean hasIndexableAttribute = hasIndexableAttribute(existingDef);
+
+            if (!hasIndexableAttribute) {
```
sheetalshah1007: The hasIndexableAttribute check could be expensive if there are many attributes. 
   Consider caching this in the BusinessMetadataDef or TypeRegistry.

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasBusinessMetadataDefStoreV2.java
```diff
@@ -271,15 +285,29 @@ public AtlasVertex preDeleteByName(String name) throws AtlasBaseException {
             throw new AtlasBaseException(AtlasErrorCode.TYPE_NAME_NOT_FOUND, name);
         }
 
-        checkBusinessMetadataRef(existingDef.getName());
+        if (!forceDelete) {
+            boolean hasIndexableAttribute = hasIndexableAttribute(existingDef);
 
-        LOG.debug("<== AtlasBusinessMetadataDefStoreV2.preDeleteByName({}): {}", name, ret);
+            if (!hasIndexableAttribute) {
```
sheetalshah1007: The hasIndexableAttribute check could be expensive if there are many attributes. 
   Consider caching this in the BusinessMetadataDef or TypeRegistry.

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasBusinessMetadataDefStoreV2.java
```diff
@@ -258,8 +266,14 @@ public AtlasBusinessMetadataDef updateByGuid(String guid, AtlasBusinessMetadataD
...
     public AtlasVertex preDeleteByName(String name) throws AtlasBaseException {
-        LOG.debug("==> AtlasBusinessMetadataDefStoreV2.preDeleteByName({})", name);
+        return preDeleteByName(name, false);
+    }
+
+    @Override
+    public AtlasVertex preDeleteByName(String name, boolean forceDelete) throws AtlasBaseException {
+        LOG.debug("==> AtlasBusinessMetadataDefStoreV2.preDeleteByName({}, {})", name, forceDelete);
 
         AtlasBusinessMetadataDef existingDef = typeRegistry.getBusinessMetadataDefByName(name);
```
sheetalshah1007, line 270: Add a warning log when force-delete is used; something like "Force-deleting BusinessMetadata '{}'. Skipping validation - orphaned references may remain.", name"

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasBusinessMetadataDefStoreV2.java
```diff
@@ -271,15 +285,29 @@ public AtlasVertex preDeleteByName(String name) throws AtlasBaseException {
...
     public AtlasVertex preDeleteByGuid(String guid) throws AtlasBaseException {
-        LOG.debug("==> AtlasBusinessMetadataDefStoreV2.preDeleteByGuid({})", guid);
+        return preDeleteByGuid(guid, false);
+    }
+
+    @Override
+    public AtlasVertex preDeleteByGuid(String guid, boolean forceDelete) throws AtlasBaseException {
+        LOG.debug("==> AtlasBusinessMetadataDefStoreV2.preDeleteByGuid({}, {})", guid, forceDelete);
 
         AtlasBusinessMetadataDef existingDef = typeRegistry.getBusinessMetadataDefByGuid(guid);
```
sheetalshah1007, line 296: Add a warning log when force-delete is used; something like "Force-deleting BusinessMetadata '{}'. Skipping validation - orphaned references may remain.", name"

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasBusinessMetadataDefStoreV2.java
```diff
@@ -271,15 +285,29 @@ public AtlasVertex preDeleteByName(String name) throws AtlasBaseException {
             throw new AtlasBaseException(AtlasErrorCode.TYPE_NAME_NOT_FOUND, name);
         }
 
-        checkBusinessMetadataRef(existingDef.getName());
+        if (!forceDelete) {
```
sheetalshah1007: The logic at lines 288-296 and 322-330 (in preDeleteByGuid) is duplicated. 
   Extract to a helper method: validateDeletion(AtlasBusinessMetadataDef def, boolean forceDelete)

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasBusinessMetadataDefStoreV2.java
```diff
@@ -291,15 +319,35 @@ public AtlasVertex preDeleteByGuid(String guid) throws AtlasBaseException {
...
         return ret;
     }
 
+    private boolean hasIndexableAttribute(AtlasBusinessMetadataDef bmDef) {
+        if (bmDef == null || CollectionUtils.isEmpty(bmDef.getAttributeDefs())) {
+            return false;
+        }
+
+        for (AtlasStructDef.AtlasAttributeDef attributeDef : bmDef.getAttributeDefs()) {
+            if (attributeDef.getIsIndexable()) {
```
sheetalshah1007: ****CRITICAL**: Logic is inverted - checks for good attributes instead of problematic ones.**

Example: BM with attr1 (String, indexable) + attr2 (Array, non-indexable)
- Current: hasIndexableAttribute() = true → normal delete allowed → SLOW (still checks array)
- Should be: **hasNonIndexableAttribute() = true → force-delete required**

The check should block normal deletion if ANY attribute is non-indexable, not only if ALL attributes are non-indexable.

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasBusinessMetadataDefStoreV2.java
```diff
@@ -363,44 +411,91 @@ private AtlasBusinessMetadataDef toBusinessMetadataDef(AtlasVertex vertex) throw
...
+
+        boolean isPresent = isBusinessAttributePresentInGraph(vertexPropertyName, allApplicableTypes);
+
+        if (LOG.isDebugEnabled()) {
+            LOG.info("Reference check for attribute {} took {} ms. Found: {}",
+                    attributeDef.getName(), (System.currentTimeMillis() - startTime), isPresent);
+        }
+
+        if (isPresent) {
+            throw new AtlasBaseException(AtlasErrorCode.TYPE_HAS_REFERENCES, bmDef.getName());
```
sheetalshah1007, line 448: Current problem: Error "TYPE_HAS_REFERENCES" doesn't say which attribute is blocking deletion or where to look.

Suggestion: Add detailed logging before throwing the error for easier debugging. 
```
LOG.error("Cannot delete BusinessMetadata '{}' - attribute '{}' (vertex property: '{}') 
has references in entity types: {}", bmDef.getName(), attributeDef.getName(), 
vertexPropertyName, allApplicableTypes);

```
This way logs would show exactly which attribute, property name, and entity types have references. 
Using this information, user can run a targeted query and find the exact references

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasBusinessMetadataDefStoreV2.java
```diff
@@ -363,44 +411,91 @@ private AtlasBusinessMetadataDef toBusinessMetadataDef(AtlasVertex vertex) throw
...
 
-                if (CollectionUtils.isNotEmpty(applicableTypes) && isBusinessAttributePresent(vertexPropertyName, applicableTypes)) {
-                    throw new AtlasBaseException(AtlasErrorCode.TYPE_HAS_REFERENCES, typeName);
-                }
-            }
+        if (CollectionUtils.isEmpty(applicableTypes)) {
+            return;
+        }
+
+        Set<String> allApplicableTypes = getApplicableTypesWithSubTypes(applicableTypes);
```
sheetalshah1007: `getApplicableTypesWithSubTypes() `gets called multiple times for the same BM definition.

Optimization: 
Calculate `allApplicableTypes` once at the BM level and pass it down:
- Move `getApplicableTypesWithSubTypes` call to `checkBusinessMetadataRef`
- Pass the expanded types to `validateAttributeReferences`

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasBusinessMetadataDefStoreV2.java
```diff
@@ -363,44 +411,91 @@ private AtlasBusinessMetadataDef toBusinessMetadataDef(AtlasVertex vertex) throw
...
 
-    private boolean isBusinessAttributePresent(String attrName, Set<String> applicableTypes) throws AtlasBaseException {
-        SearchParameters.FilterCriteria criteria = new SearchParameters.FilterCriteria();
 
-        criteria.setAttributeName(attrName);
-        criteria.setOperator(SearchParameters.Operator.NOT_EMPTY);
+    private boolean isBusinessAttributePresentInGraph(String vertexPropertyName, Set<String> allApplicableTypes) {
+        if (graph == null || CollectionUtils.isEmpty(allApplicableTypes)) {
+            return false;
+        }
```
sheetalshah1007: **CRITICAL**: The null graph check returns false, meaning "no references exist, safe to delete."

This is dangerous! If the graph is unavailable (`null`), we should fail-safe by blocking 
deletion, not allowing it.

Change to:
```
if (graph == null) {
    throw new IllegalStateException("Graph instance is not available - cannot verify references");
}
if (CollectionUtils.isEmpty(allApplicableTypes)) {
    return false; // This is fine - no types to check
}
```

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasBusinessMetadataDefStoreV2.java
```diff
@@ -363,44 +411,91 @@ private AtlasBusinessMetadataDef toBusinessMetadataDef(AtlasVertex vertex) throw
...
+        try {
+            List<String> typesList = new ArrayList<>(allApplicableTypes);
 
-        entityFilters.setCondition(SearchParameters.FilterCriteria.Condition.OR);
-        entityFilters.setCriterion(Collections.singletonList(criteria));
+            // 1. To Check if the BM property exists on a vertex where the direct type matches
+            Iterable<AtlasVertex> verticesDirect = graph.query()
+                    .has(vertexPropertyName, AtlasGraphQuery.ComparisionOperator.NOT_EQUAL, (Object) null)
+                    .in(Constants.ENTITY_TYPE_PROPERTY_KEY, typesList)
+                    .vertices();
```
sheetalshah1007: Consider adding a limit (e.g., .limit(1)) since we only need to know if ANY exist

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasBusinessMetadataDefStoreV2.java
```diff
@@ -363,44 +411,91 @@ private AtlasBusinessMetadataDef toBusinessMetadataDef(AtlasVertex vertex) throw
...
-        searchParameters.setEntityFilters(entityFilters);
-        searchParameters.setAttributes(Collections.singleton(attrName));
-        searchParameters.setOffset(0);
-        searchParameters.setLimit(1);
+            // 2. To Check if the BM property exists on a vertex where it inherits from one of parent Types
+            // This is crucial for Case 6 (Parent -> Child)
+            Iterable<AtlasVertex> verticesInherited = graph.query()
+                    .has(vertexPropertyName, AtlasGraphQuery.ComparisionOperator.NOT_EQUAL, (Object) null)
+                    .in(Constants.SUPER_TYPES_PROPERTY_KEY, typesList)
+                    .vertices();
```
sheetalshah1007: Consider adding a limit (e.g., .limit(1)) since we only need to know if ANY exist

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasBusinessMetadataDefStoreV2.java
```diff
@@ -363,44 +411,91 @@ private AtlasBusinessMetadataDef toBusinessMetadataDef(AtlasVertex vertex) throw
...
-        AtlasSearchResult atlasSearchResult = entityDiscoveryService.searchWithParameters(searchParameters);
+            if (verticesInherited != null && verticesInherited.iterator().hasNext()) {
+                return true;
+            }
+        } catch (Exception e) {
+            LOG.error("Error occurred while querying graph for references of property: {}", vertexPropertyName, e);
+            return true;
+        }
+        return false;
+    }
```
sheetalshah1007, line 479: The two queries (for direct and inherited type match) could be combined into one to make it more efficient

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasBusinessMetadataDefStoreV2.java
```diff
@@ -363,44 +411,91 @@ private AtlasBusinessMetadataDef toBusinessMetadataDef(AtlasVertex vertex) throw
...
+            Iterable<AtlasVertex> verticesInherited = graph.query()
+                    .has(vertexPropertyName, AtlasGraphQuery.ComparisionOperator.NOT_EQUAL, (Object) null)
+                    .in(Constants.SUPER_TYPES_PROPERTY_KEY, typesList)
+                    .vertices();
 
-        AtlasSearchResult atlasSearchResult = entityDiscoveryService.searchWithParameters(searchParameters);
+            if (verticesInherited != null && verticesInherited.iterator().hasNext()) {
+                return true;
+            }
+        } catch (Exception e) {
```
sheetalshah1007, line 474: **CRITICAL:** Catching all exceptions and returning true treats temporary system errors (network issues, DB timeouts) the same as actual reference blockers. Any temporary failure makes deletion fail with "`TYPE_HAS_REFERENCES`" error - even when no references exist.  Users can't tell if they need to clean up references or just retry. Throw an exception with a clear "system error" message instead of silently returning true.

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasBusinessMetadataDefStoreV2.java
```diff
@@ -363,44 +411,91 @@ private AtlasBusinessMetadataDef toBusinessMetadataDef(AtlasVertex vertex) throw
...
+            LOG.error("Error occurred while querying graph for references of property: {}", vertexPropertyName, e);
+            return true;
+        }
+        return false;
+    }
+
+    private Set<String> getApplicableTypesWithSubTypes(Set<String> applicableTypes) throws AtlasBaseException {
```
sheetalshah1007: Suggestions:
1. Add JavaDoc explaining WHY sub-types are needed as this is an important business logic -  without JavaDoc, developers won't understand the inheritance scenario.

2. Consider caching the result in TypeRegistry to avoid repeated calculations

webapp/src/test/java/org/apache/atlas/web/rest/TypesRESTTest.java
sheetalshah1007, line 1: Missing test coverage: Test the actual use case - non-indexable BM deletion

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasBusinessMetadataDefStoreV2.java
```diff
@@ -363,44 +411,91 @@ private AtlasBusinessMetadataDef toBusinessMetadataDef(AtlasVertex vertex) throw
...
         }
     }
 
-    private boolean isBusinessAttributePresent(String attrName, Set<String> applicableTypes) throws AtlasBaseException {
-        SearchParameters.FilterCriteria criteria = new SearchParameters.FilterCriteria();
 
-        criteria.setAttributeName(attrName);
-        criteria.setOperator(SearchParameters.Operator.NOT_EMPTY);
+    private boolean isBusinessAttributePresentInGraph(String vertexPropertyName, Set<String> allApplicableTypes) {
+        if (graph == null || CollectionUtils.isEmpty(allApplicableTypes)) {
```
sheetalshah1007: **CRITICAL**: Remove the graph null check at line 451. Returning false when graph is null means "safe to delete" without actually verifying references - this is dangerous. AtlasGraph is dependency-injected and cannot be null in production.

repository/src/main/java/org/apache/atlas/repository/store/graph/AtlasTypeDefGraphStore.java
```diff
@@ -539,14 +539,20 @@ public AtlasTypesDef createUpdateTypesDef(AtlasTypesDef typesToCreate, AtlasType
     @Override
     @GraphTransaction
     public void deleteTypesDef(AtlasTypesDef typesDef) throws AtlasBaseException {
+        deleteTypesDef(typesDef, false);
+    }
+
+    @GraphTransaction
+    public void deleteTypesDef(AtlasTypesDef typesDef, boolean forceDelete) throws AtlasBaseException {
         if (LOG.isDebugEnabled()) {
```
sheetalshah1007: LOG.isDebugEnabled() check can be removed

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasBusinessMetadataDefStoreV2.java
```diff
@@ -363,44 +411,91 @@ private AtlasBusinessMetadataDef toBusinessMetadataDef(AtlasVertex vertex) throw
...
+        String qualifiedName      = AtlasStructType.AtlasAttribute.getQualifiedAttributeName(bmDef, attributeDef.getName());
+        String vertexPropertyName = AtlasStructType.AtlasAttribute.generateVertexPropertyName(bmDef, attributeDef, qualifiedName);
+
+        long startTime = System.currentTimeMillis();
+
+        boolean isPresent = isBusinessAttributePresentInGraph(vertexPropertyName, allApplicableTypes);
+
+        if (LOG.isDebugEnabled()) {
+            LOG.info("Reference check for attribute {} took {} ms. Found: {}",
+                    attributeDef.getName(), (System.currentTimeMillis() - startTime), isPresent);
```
sheetalshah1007: Code/logging related to time capture can be removed

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasTypeDefGraphStoreV2.java
```diff
@@ -73,14 +73,11 @@ public class AtlasTypeDefGraphStoreV2 extends AtlasTypeDefGraphStore {
 
     protected final AtlasGraph atlasGraph;
 
-    private final EntityDiscoveryService entityDiscoveryService;
-
     @Inject
     public AtlasTypeDefGraphStoreV2(AtlasTypeRegistry typeRegistry, List<TypeDefChangeListener> typeDefChangeListeners, AtlasGraph atlasGraph, EntityDiscoveryService entityDiscoveryService) {
```
sheetalshah1007: The constructor still takes `EntityDiscoveryService entityDiscoveryService `but no longer stores or uses it:
Remove the parameter to fix the unwanted wiring

webapp/src/main/java/org/apache/atlas/web/rest/TypesREST.java
sheetalshah1007: **Bulk REST delete has no force**
`DELETE` `/api/atlas/v2/types/typedefs` still calls `typeDefStore.deleteTypesDef(typesDef)` (always `forceDelete = false` internally). **Single-name** delete supports `?force=true`, but **bulk** does not. 

Suggestion: add an optional force query param to `deleteAtlasTypeDefs` and align bulk delete with `force`

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasBusinessMetadataDefStoreV2.java
```diff
@@ -360,47 +405,104 @@ private AtlasBusinessMetadataDef toBusinessMetadataDef(AtlasVertex vertex) throw
...
+                AtlasBusinessMetadataType.AtlasBusinessAttribute bmAttr =
+                        (AtlasBusinessMetadataType.AtlasBusinessAttribute) bmType.getAttribute(attributeDef.getName());
+                Set<String> cachedTypes = bmAttr != null ? bmAttr.getApplicableEntityTypeNamesWithSubTypes() : Collections.emptySet();
+
+                allApplicableTypes = CollectionUtils.isNotEmpty(cachedTypes) ? cachedTypes : getApplicableTypesWithSubTypes(attributeDef);
+            } else {
+                allApplicableTypes = getApplicableTypesWithSubTypes(attributeDef);
             }
+
+            LOG.info("REF-CHECK [ApplicableTypes] BM='{}' attr='{}' types={}",
```
sheetalshah1007: logs **INFO** for every attribute's applicable types
Suggestion: prefer DEBUG logging for this

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasBusinessMetadataDefStoreV2.java
```diff
@@ -360,47 +405,104 @@ private AtlasBusinessMetadataDef toBusinessMetadataDef(AtlasVertex vertex) throw
...
+    private void validateAttributeReferences(AtlasBusinessMetadataDef bmDef, AtlasStructDef.AtlasAttributeDef attributeDef,
+                                             Set<String> allApplicableTypes, String identifier) throws AtlasBaseException {
+        String qualifiedName      = AtlasStructType.AtlasAttribute.getQualifiedAttributeName(bmDef, attributeDef.getName());
+        String vertexPropertyName = AtlasStructType.AtlasAttribute.generateVertexPropertyName(bmDef, attributeDef, qualifiedName);
+        boolean isPresent         = isBusinessAttributePresentInGraph(vertexPropertyName, allApplicableTypes);
 
-        criteria.setAttributeName(attrName);
-        criteria.setOperator(SearchParameters.Operator.NOT_EMPTY);
+        if (isPresent) {
+            LOG.error("Cannot delete BusinessMetadata '{}' (request='{}') - attribute '{}' (vertex property: '{}') has references in entity types: {}",
```
sheetalshah1007: Logging ERROR for existing references is misleading. Use **WARN** (or INFO) and reserve ERROR for real failures.

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasTypeDefGraphStoreV2.java
```diff
@@ -73,14 +73,11 @@ public class AtlasTypeDefGraphStoreV2 extends AtlasTypeDefGraphStore {
...
         super(typeRegistry, typeDefChangeListeners);
 
-        this.atlasGraph             = atlasGraph;
-        this.entityDiscoveryService = entityDiscoveryService;
+        this.atlasGraph = atlasGraph;
 
         LOG.debug("<== AtlasTypeDefGraphStoreV1()");
```
sheetalshah1007: AtlasTypeDefGraphStoreV2 still logs "AtlasTypeDefGraphStoreV1()"

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasBusinessMetadataDefStoreV2.java
```diff
@@ -360,47 +405,104 @@ private AtlasBusinessMetadataDef toBusinessMetadataDef(AtlasVertex vertex) throw
...
+                    .has(vertexPropertyName, AtlasGraphQuery.ComparisionOperator.NOT_EQUAL, (Object) null);
 
-        searchParameters.setTypeName(String.join(SearchContext.TYPENAME_DELIMITER, applicableTypes));
-        searchParameters.setExcludeDeletedEntities(true);
-        searchParameters.setIncludeSubClassifications(false);
-        searchParameters.setEntityFilters(entityFilters);
-        searchParameters.setAttributes(Collections.singleton(attrName));
-        searchParameters.setOffset(0);
-        searchParameters.setLimit(1);
+            List<AtlasGraphQuery> orConditions = new ArrayList<>();
```
sheetalshah1007, line 463: Remove the inline comments here—the code is clear enough without them

sheetalshah1007 (changes requested):

intg/src/main/java/org/apache/atlas/AtlasErrorCode.java
```diff
@@ -259,7 +259,9 @@ public enum AtlasErrorCode {
     IMPORT_REGISTRATION_FAILED(500, "ATLAS-500-00-020", "Failed to register import request"),
     IMPORT_FAILED(500, "ATLAS-500-00-021", "Import with id {0} failed"),
     ABORT_IMPORT_FAILED(500, "ATLAS-500-00-022", "Failed to abort import with id {0}"),
-    IMPORT_QUEUEING_FAILED(500, "ATLAS-500-00-023", "Failed to add import with id {0} to request queue, please try again later");
+    IMPORT_QUEUEING_FAILED(500, "ATLAS-500-00-023", "Failed to add import with id {0} to request queue, please try again later"),
+
+    NON_INDEXABLE_BM_DELETE_NOT_ALLOWED(400, "ATLAS-400-00-106", "Deletion not allowed for non-indexable Business Metadata ''{0}'' without force=true. Non-indexable attributes cannot be validated efficiently for references; use force=true to skip validation and delete (warning: orphaned references may remain).");
```
sheetalshah1007: **`ATLAS-400-00-106 `is already used** by `INVALID_RELATIONSHIP_LABEL`. Please move `NON_INDEXABLE_BM_DELETE_NOT_ALLOWED` into the `400` section and use` ATLAS-400-00-107` (or the **next free** 400 id), not a duplicate 106.

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasTypeDefGraphStoreV2.java
sheetalshah1007, line 181: `AtlasTypeDefGraphStoreV2.init()` — change log messages from AtlasTypeDefGraphStoreV1 to V2

webapp/src/main/java/org/apache/atlas/web/rest/TypesREST.java
```diff
@@ -405,19 +407,24 @@ public AtlasTypesDef updateAtlasTypeDefs(final AtlasTypesDef typesDef) throws At
      * @HTTP 204 On successful deletion of the requested type definitions
      * @HTTP 400 On validation failure for any type definitions
      */
+    public void deleteAtlasTypeDefs(final AtlasTypesDef typesDef) throws AtlasBaseException {
```
sheetalshah1007: Consider keeping only the `@DELETE` bulk-delete method (with `force`), and drop the extra overload that only passes `false`, so tests and Java callers use `deleteAtlasTypeDefs(typesDef, false) `explicitly—same idea as delete-by-name.

webapp/src/main/java/org/apache/atlas/web/rest/TypesREST.java
```diff
@@ -405,19 +407,24 @@ public AtlasTypesDef updateAtlasTypeDefs(final AtlasTypesDef typesDef) throws At
...
+        deleteAtlasTypeDefs(typesDef, false);
+    }
+
     @DELETE
     @Path("/typedefs")
     @Experimental
     @Timed
-    public void deleteAtlasTypeDefs(final AtlasTypesDef typesDef) throws AtlasBaseException {
+    public void deleteAtlasTypeDefs(final AtlasTypesDef typesDef,
```
sheetalshah1007, line 423: Add JavaDoc to `deleteAtlasTypeDefs` (the @`DELETE` method with `force`) —` @param forceDelete` and BM semantics match `deleteAtlasTypeByName`; the REST endpoint currently has no class-level docs for force.

sheetalshah1007 (approved):

webapp/src/main/java/org/apache/atlas/web/rest/TypesREST.java
```diff
@@ -399,25 +401,34 @@ public AtlasTypesDef updateAtlasTypeDefs(final AtlasTypesDef typesDef) throws At
     }
 
     /**
-     * Bulk delete API for all types
-     * @param typesDef A composite object that captures all types to be deleted
+     * Bulk delete API for all types.
+     *
+     * @param typesDef A composite object that captures all types to be deleted.
+     * @param forceDelete If true, bypasses pre-delete validation checks and forcefully removes type definitions.
```
pinal-shah: Please mention, only for businessMetadata typdef

webapp/src/main/java/org/apache/atlas/web/rest/TypesREST.java
```diff
@@ -426,22 +437,30 @@ public void deleteAtlasTypeDefs(final AtlasTypesDef typesDef) throws AtlasBaseEx
     /**
      * Delete API for type identified by its name.
      * @param typeName Name of the type to be deleted.
+     * @param forceDelete If true, bypasses pre-delete validation checks and forcefully removes the type definition.
```
pinal-shah: Please add, only supported for businessMetadata

repository/src/main/java/org/apache/atlas/repository/store/graph/AtlasDefStore.java
```diff
@@ -45,9 +49,33 @@ public interface AtlasDefStore<T extends AtlasBaseTypeDef> {
...
-
     AtlasVertex preDeleteByGuid(String guid) throws AtlasBaseException;
 
+    void deleteByName(String name, AtlasVertex preDeleteResult) throws AtlasBaseException;
+
     void deleteByGuid(String guid, AtlasVertex preDeleteResult) throws AtlasBaseException;
+
+    default AtlasVertex preDeleteByName(String name, boolean forceDelete) throws AtlasBaseException {
+        if (forceDelete) {
+            LOG.debug("Force-delete flag ignored in {}.preDeleteByName() for type '{}'; feature is implemented only for BusinessMetadata.", getClass().getSimpleName(), name);
```
pinal-shah: no need to log here

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasAbstractDefStoreV2.java
```diff
@@ -148,24 +148,46 @@ public boolean isValidName(String typeName) {
...
         AtlasVertex vertex = (preDeleteResult == null) ? preDeleteByName(name) : preDeleteResult;
 
         typeDefStore.deleteTypeVertex(vertex);
 
-        LOG.debug("<== AtlasAbstractDefStoreV1.deleteByName({}, {})", name, preDeleteResult);
+        LOG.debug("<== AtlasAbstractDefStoreV2.deleteByName({}, {})", name, preDeleteResult);
     }
 
     @Override
     public void deleteByGuid(String guid, AtlasVertex preDeleteResult) throws AtlasBaseException {
```
pinal-shah, line 155: @Override
    public void deleteByGuid(String guid, AtlasVertex preDeleteResult) throws AtlasBaseException {
        deleteByGuid(guid, preDeleteResult, false);
    }
Please review above

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasBusinessMetadataDefStoreV2.java
```diff
@@ -360,47 +405,100 @@ private AtlasBusinessMetadataDef toBusinessMetadataDef(AtlasVertex vertex) throw
...
-                if (CollectionUtils.isNotEmpty(applicableTypes) && isBusinessAttributePresent(vertexPropertyName, applicableTypes)) {
-                    throw new AtlasBaseException(AtlasErrorCode.TYPE_HAS_REFERENCES, typeName);
-                }
+            if (bmType != null) {
+                AtlasBusinessMetadataType.AtlasBusinessAttribute bmAttr =
+                        (AtlasBusinessMetadataType.AtlasBusinessAttribute) bmType.getAttribute(attributeDef.getName());
+                Set<String> cachedTypes = bmAttr != null ? bmAttr.getApplicableEntityTypeNamesWithSubTypes() : Collections.emptySet();
+
+                allApplicableTypes = CollectionUtils.isNotEmpty(cachedTypes) ? cachedTypes : getApplicableTypesWithSubTypes(attributeDef);
+            } else {
```
pinal-shah: if bmType == null, it shouldn't proceed for anything

intg/src/main/java/org/apache/atlas/type/AtlasBusinessMetadataType.java
```diff
@@ -137,6 +151,29 @@ void resolveReferencesPhase2(AtlasTypeRegistry typeRegistry) throws AtlasBaseExc
         }
     }
 
+    @Override
+    void resolveReferencesPhase3(AtlasTypeRegistry typeRegistry) throws AtlasBaseException {
```
pinal-shah, line 155: please review if this can be achieved in resolveReferencesPhase2

intg/src/main/java/org/apache/atlas/type/AtlasBusinessMetadataType.java
```diff
@@ -115,10 +123,16 @@ void resolveReferences(AtlasTypeRegistry typeRegistry) throws AtlasBaseException
                 bmAttribute = new AtlasBusinessAttribute(attribute, entityTypes);
             }
 
+            if (!hasNonIndexableAttributes && !Boolean.TRUE.equals(attributeDef.getIsIndexable())) {
```
pinal-shah: Simplify !Boolean.TRUE.equals(attributeDef.getIsIndexable() to -> !attributeDef.getIsIndexable()

Please incoporate other places as well

intg/src/main/java/org/apache/atlas/AtlasErrorCode.java
```diff
@@ -183,6 +183,7 @@ public enum AtlasErrorCode {
     BLANK_NAME_ATTRIBUTE(400, "ATLAS-400-00-104", "Name Attribute can't be empty!"),
     BLANK_VALUE_ATTRIBUTE(400, "ATLAS-400-00-105", "Value Attribute can't be empty!"),
     INVALID_RELATIONSHIP_LABEL(400, "ATLAS-400-00-106", "Invalid relationship label {0}. The referenced entity type {1} could not be resolved from the type registry."),
+    NON_INDEXABLE_BM_DELETE_NOT_ALLOWED(400, "ATLAS-400-00-107", "Deletion not allowed for non-indexable Business Metadata ''{0}'' without force=true. Non-indexable attributes cannot be validated efficiently for references; use force=true to skip validation and delete (warning: orphaned references may remain)."),
```
pinal-shah, line 186: What do you mean by orphaned references may remain?

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasBusinessMetadataDefStoreV2.java
```diff
@@ -291,15 +302,49 @@ public AtlasVertex preDeleteByGuid(String guid) throws AtlasBaseException {
...
+    }
+
+    private boolean hasNonIndexableAttribute(AtlasBusinessMetadataDef businessMetadataDef) {
+        AtlasBusinessMetadataType bmType = typeRegistry.getBusinessMetadataTypeByName(businessMetadataDef.getName());
+
+        if (bmType != null && bmType.hasNonIndexableAttributes()) {
+            return true;
+        }
+
+        if (CollectionUtils.isNotEmpty(businessMetadataDef.getAttributeDefs())) {
```
pinal-shah, line 337: please review if this is required #337 to #343

pinal-shah (changes requested):

intg/src/main/java/org/apache/atlas/type/AtlasBusinessMetadataType.java
UmeshPatil-1: We kept this in `resolveReferencesPhase3` because it expands applicable entity types using `getAllSubTypes()`, which is only fully populated after all entity types have finished phase 2. Moving it into BM’s phase 2 could run before some entities have registered subtypes, so expansion could be incomplete.

intg/src/main/java/org/apache/atlas/AtlasErrorCode.java
UmeshPatil-1: With force=true, Atlas deletes the typedef without checking whether entities still hold values for those attributes, such values may remain on entity instances until separately cleaned up.

pinal-shah (approved):

PR #511, opened 2026-02-02
ATLAS-5159:- Export APIs: Exporting parent type doesn't export child type

UmeshPatil-1:
**ATLAS-5159**

**Background:**
Exporting entities of a parent type (e.g., Referenceable) does not export entities of its child type (e.g., hive_table).
When attempting to export using the parent type with unique attributes, no child entities were returned, leading to incomplete exports.

**Root Cause:**

The previous implementation of StartEntityFetchByExportRequest did not consistently account for the type hierarchy. Specifically, when fetching by uniqueAttributes/matchType, the logic did not automatically include subtypes, meaning only entities explicitly labeled with the parent type were discovered.

**Changes Proposed:**

Hierarchical Type Lookup: Updated getEntitiesForMatchTypeUsingUniqueAttributes to identify the AtlasEntityType and automatically include all its subtypes (entityType.getAllSubTypes()) in the search criteria.

Gremlin Binding Expansion: Refactored getBindingsForTypeName to handle comma-separated type names and recursively resolve all subtypes for each provided type. This ensures that a query for a parent type translates into a Gremlin search across the entire inheritance tree.

Enhanced Match Logic:
Improved getEntitiesForMatchTypeType to utilize the expanded type list.

**Impact:**

Exporting parent types like Referenceable now correctly includes all child entities.
Maintains consistent behavior for all fetch types (FULL, INCREMENTAL, CONNECTED).
Minimal impact on performance; no additional schema changes are required.

**How was this patch tested?**

Maven Build:- Build Successfull.
Manual testing of export requests with all three fetch types: FULL, INCREMENTAL, CONNECTED.
Verified entity GUIDs are correctly returned for parent types and child entities,Concrete types with unique attributes (hive_table)
Confirmed behavior for concrete types, abstract types, and invalid/misspelled types.
Logs verified query execution, fetch type handling, and error reporting.


repository/src/main/java/org/apache/atlas/repository/impexp/StartEntityFetchByExportRequest.java
```diff
@@ -109,6 +109,11 @@ public List<String> get(AtlasExportRequest exportRequest, AtlasObjectId item) {
 
                 return ret;
             }
+
+            if (StringUtils.isEmpty(item.getTypeName()) && MapUtils.isNotEmpty(item.getUniqueAttributes())) {
+                LOG.info("Request missing typeName but has uniqueAttributes. Attempting generic type search.");
+                return getEntitiesForMatchTypeType(item, MATCH_TYPE_FOR_TYPE);
+            }
```
chaitalicod, line 118: NPE in get() finally block. getEntitiesForMatchTypeType → executeGremlinQuery returns null on ScriptException. The new fallback assigns that to ret and returns, triggering finally → ret.size() → NPE (also masks the real error). Null-guard the log, or return an empty list instead of null from executeGremlinQuery.

repository/src/main/java/org/apache/atlas/repository/impexp/StartEntityFetchByExportRequest.java
```diff
@@ -165,27 +176,49 @@ private List<String> getEntitiesForMatchTypeUsingUniqueAttributes(AtlasObjectId
...
-    private HashMap<String, Object> getBindingsForTypeName(String typeName) {
-        HashMap<String, Object> ret = new HashMap<>();
-
-        ret.put(BINDING_PARAMETER_TYPENAME, new HashSet<>(Arrays.asList(StringUtils.split(typeName, ","))));
+    private HashMap<String, Object> getBindingsForTypeName(String typeName) throws AtlasBaseException {
+        HashMap<String, Object> ret              = new HashMap<>();
+        Set<String>             typeNamesToQuery = new HashSet<>();
+
+        if (StringUtils.isBlank(typeName)) {
+            typeNamesToQuery.addAll(typeRegistry.getAllEntityDefNames());
```
chaitalicod, line 239: Regression: no-typeName fallback ignores uniqueAttributes. The fallback routes to getEntitiesForMatchTypeType, which runs EXPORT_TYPE_ALL_FOR_TYPE over all entity types and drops the attribute filter. A request with only uniqueAttributes now matches every entity in the system, and the caller picks guids.get(0) (arbitrary). 

Please check the impact on other usecases and regression if its fine

PR #640, opened 2026-05-21
ATLAS-5190: Upgrading jersey version to fix the conflicts caused by jackson upgrade

aditya-gupta36:
## How was this patch tested?

- Build - Docker Build, Local Build
- UI Test, Curl API testing, login/logout, start/stop/quick_start scripts, hive query through beeline @docker env

- build passed: mvn -Pdist,embedded-solr-it -DskipTests=false -Drat.skip=true -DskipDocs clean verify

image

### DIFF
--- a/addons/falcon-bridge/pom.xml
+++ b/addons/falcon-bridge/pom.xml
@@ -54,6 +54,10 @@
                     <groupId>com.sun.jersey</groupId>
                     <artifactId>*</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>org.apache.atlas</groupId>
+                    <artifactId>atlas-client-v2-shaded</artifactId>
+                </exclusion>
             </exclusions>
         </dependency>
         <!-- Logging -->
@@ -111,6 +115,12 @@
                 </exclusion>
             </exclusions>
         </dependency>
+
+        <dependency>
+            <groupId>org.apache.atlas</groupId>
+            <artifactId>atlas-client-v2</artifactId>
+            <scope>test</scope>
+        </dependency>
         <dependency>
             <groupId>org.apache.atlas</groupId>
             <artifactId>atlas-graphdb-impls</artifactId>
@@ -126,6 +136,22 @@
 
     <build>
         <plugins>
+            <!-- Parent disables Failsafe forking (forkCount=0). Exclude shaded v2 client from IT classpath
+                 to avoid Jersey NoSuchMethodError on org.apache.atlas.shaded.javax.ws.rs.RuntimeType. -->
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-failsafe-plugin</artifactId>
+                <configuration combine.self="merge">
+                    <forkCount>1</forkCount>
+                    <reuseForks>false</reuseForks>
+                    <useSystemClassLoader>false</useSystemClassLoader>
+                    <classpathDependencyExcludes>
+                        <classpathDependencyExclude>org.apache.atlas:atlas-client-v2-shaded</classpathDependencyExclude>
+                        <classpathDependencyExclude>org.glassfish.jersey.media:jersey-media-json-jackson</classpathDependencyExclude>
+                    </classpathDependencyExcludes>
+                </configuration>
+            </plugin>
+
             <plugin>
                 <groupId>org.eclipse.jetty</groupId>
                 <artifactId>jetty-maven-plugin</artifactId>
--- a/addons/falcon-bridge/pom.xml
+++ b/addons/falcon-bridge/pom.xml
@@ -49,6 +49,12 @@
         <dependency>
             <groupId>org.apache.atlas</groupId>
             <artifactId>hive-bridge</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>com.sun.jersey</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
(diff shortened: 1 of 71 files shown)


addons/hbase-bridge/pom.xml
```diff
@@ -297,6 +415,20 @@
 
     <build>
         <plugins>
+            <!-- Parent disables Failsafe forking (forkCount=0). Fork IT JVM for stable Atlas client + Hadoop minicluster. -->
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-failsafe-plugin</artifactId>
```
pareshddevalia, line 421: What was the need here <artifactId>maven-failsafe-plugin</artifactId> ?

addons/falcon-bridge/pom.xml
```diff
@@ -120,6 +136,22 @@
 
     <build>
         <plugins>
+            <!-- Parent disables Failsafe forking (forkCount=0). Exclude shaded v2 client from IT classpath
+                 to avoid Jersey NoSuchMethodError on org.apache.atlas.shaded.javax.ws.rs.RuntimeType. -->
```
pareshddevalia, line 140: What was the need here maven-failsafe-plugin ?

addons/hbase-bridge/pom.xml
```diff
@@ -512,6 +644,16 @@
...
                                         </artifactItem>
+                                        <artifactItem>
+                                            <groupId>javax.inject</groupId>
+                                            <artifactId>javax.inject</artifactId>
+                                            <version>${javax-inject.version}</version>
+                                        </artifactItem>
+                                        <artifactItem>
+                                            <groupId>javax.annotation</groupId>
+                                            <artifactId>javax.annotation-api</artifactId>
+                                            <version>1.3.2</version>
```
pareshddevalia, line 655: can we keep the version in pom.xml files

addons/hbase-bridge/pom.xml
```diff
@@ -582,20 +724,61 @@
                                             <artifactId>commons-configuration2</artifactId>
                                             <version>${commons-conf2.version}</version>
                                         </artifactItem>
+                                        <!-- Hadoop metrics/UGI still uses commons-configuration 1.x API -->
+                                        <artifactItem>
+                                            <groupId>commons-configuration</groupId>
+                                            <artifactId>commons-configuration</artifactId>
+                                            <version>1.10</version>
```
pareshddevalia, line 731: common-configuration 1.10 is need? we have already upgraded to 2

addons/impala-bridge/pom.xml
```diff
@@ -227,9 +284,31 @@
...
+            <version>${jersey2.version}</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
     <build>
         <plugins>
+            <!-- Parent disables Failsafe forking (forkCount=0). Fork IT JVM; on Java 9+ use
+                 HiveITSystemClassLoader so Hive 3.1.x SessionState sees a URLClassLoader. -->
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
```
pareshddevalia, line 300: Why these plugin is needed?

addons/impala-bridge/pom.xml
```diff
@@ -543,5 +627,22 @@
...
+            <id>hive-it-jdk17</id>
+            <activation>
+                <jdk>[9,)</jdk>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-failsafe-plugin</artifactId>
+                        <configuration combine.self="merge">
```
pareshddevalia, line 640: Same as here

addons/hbase-bridge/pom.xml
aditya-gupta36: Why maven-failsafe-plugin?
The parent POM sets forkCount=0, so all ITs run inside the Maven JVM with one mixed classpath. After the Jersey 2 upgrade, bridge ITs hit classpath conflicts: shaded atlas-client-v2-shaded (relocated JAX-RS → NoSuchMethodError), duplicate Jackson providers, and Hive changing the thread context class loader during SessionState.start().

We override Failsafe per module with forkCount=1, reuseForks=false, and targeted classpathDependencyExcludes. Without these overrides, CI shows failures like LinkageError on ClientBuilder, shaded...RuntimeType NoSuchMethodError, and (on JDK 17) AppClassLoader cannot be cast to URLClassLoader in Hive ITs.

Classloader issue — why it exists?
"Class loader casting to URLClassLoader succeeds in Java 8 but fails in Java 9+ due to the internal modularity changes; the modern application class loader (jdk.internal.loader.ClassLoaders$AppClassLoader) no longer extends URLClassLoader."

Why HiveITSystemClassLoader.java was added?
A custom system class loader installed at JVM startup so Hive’s SessionState.class.getClassLoader() returns a real URLClassLoader:

ITs also call SessionState.start() via the same Hive 3.1.3 code path. On JDK 17, Hive unconditionally casts SessionState.class.getClassLoader() to URLClassLoader. The default JDK 9+ app loader is not a URLClassLoader, so @BeforeClass setUp() fails before any test runs.

HiveITSystemClassLoader is a test-only system class loader (via -Djava.system.class.loader) that makes that cast succeed. It is activated only on JDK 9+ through the hive-it-jdk17 profile and requires forkCount=1.

PR #662, opened 2026-06-04
ATLAS-5310: Export/Import API : When tag attributes types get modified between th…

saksenasonali:
Fixes Replication Manager export/import when a classification (tag) attribute type changes between exports (e.g. string → double after the tag is deleted and recreated on the source with the same name). Previously, incremental import failed with INVALID_IMPORT_ATTRIBUTE_TYPE_CHANGED or a JanusGraph schema error because the target kept the old typedef and property key types.

Changes:

TypeAttributeDifference — When an attribute already exists but its typeName differs, treat the incoming definition as an update (via addAttribute) instead of throwing INVALID_IMPORT_ATTRIBUTE_TYPE_CHANGED.

AtlasStructDefStoreV2 — During import (RequestContext.isImportInProgress()), allow persisting typedef attribute data-type changes and serialize the new attribute definition from the import payload.

GraphBackedSearchIndexer — During import, if an existing JanusGraph property key’s data type no longer matches the typedef, delete and recreate the property key so classification values can be stored with the new type.

AtlasGraphManagement / AtlasJanusGraphManagement — Add propertyKeyHasDataType() to detect property key / data-type mismatches.

How was this patch tested?
Unit / integration tests

mvn -pl repository -Dtest=TypeAttributeDifferenceTest,ImportServiceTest#importHdfs_path1 test
mvn -pl graphdb/janus -Dtest=AtlasJanusGraphManagementTest test
TypeAttributeDifferenceTest#attributeTypeChanged_ReturnsUpdatedAttribute — Verifies changed attribute types are included in the typedef diff for import.
ImportServiceTest#importHdfs_path1 — End-to-end import scenario: pre-existing tag with attrib1=date, import zip with attrib1=string; import succeeds and typedef is updated (previously expected failure).
AtlasJanusGraphManagementTest — Covers new propertyKeyHasDataType API.
Manual / cluster E2E (Replication Manager flow)

Create table + tag (a = string), associate a="abc", full incremental export (changeMarker=0), import to target
Disassociate tag, delete tag, recreate tag (a = double), associate a=3.14, incremental export, import to target
Verify target typedef has a: double and classification value 3.14

### DIFF
--- a/repository/src/main/java/org/apache/atlas/repository/impexp/TypeAttributeDifference.java
+++ b/repository/src/main/java/org/apache/atlas/repository/impexp/TypeAttributeDifference.java
@@ -18,7 +18,6 @@
 package org.apache.atlas.repository.impexp;
 
 import com.google.common.annotations.VisibleForTesting;
-import org.apache.atlas.AtlasErrorCode;
 import org.apache.atlas.exception.AtlasBaseException;
 import org.apache.atlas.model.impexp.AtlasImportResult;
 import org.apache.atlas.model.typedef.AtlasBusinessMetadataDef;
@@ -170,12 +169,11 @@ public class TypeAttributeDifference {
             if (relationshipType == null) {
                 difference.add(incoming);
             }
-        } else {
-            if (!existingAttribute.getTypeName().equals(incoming.getTypeName())) {
-                LOG.error("Attribute definition difference found: {}, {}", existingAttribute, incoming);
+        } else if (!existingAttribute.getTypeName().equals(incoming.getTypeName())) {
+            LOG.info("Attribute type changed for {}.{}: {} -> {}; updating typedef during import",
+                    existing.getName(), incoming.getName(), existingAttribute.getTypeName(), incoming.getTypeName());
 
-                throw new AtlasBaseException(AtlasErrorCode.INVALID_IMPORT_ATTRIBUTE_TYPE_CHANGED, existing.getName(), existingAttribute.getName(), existingAttribute.getTypeName(), incoming.getTypeName());
-            }
+            difference.add(incoming);
         }
     }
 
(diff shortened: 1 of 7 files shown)


repository/src/main/java/org/apache/atlas/repository/impexp/TypeAttributeDifference.java
```diff
@@ -170,12 +169,11 @@ private void updateCollectionWithDifferingAttributes(List<AtlasStructDef.AtlasAt
             if (relationshipType == null) {
                 difference.add(incoming);
             }
-        } else {
-            if (!existingAttribute.getTypeName().equals(incoming.getTypeName())) {
-                LOG.error("Attribute definition difference found: {}, {}", existingAttribute, incoming);
+        } else if (!existingAttribute.getTypeName().equals(incoming.getTypeName())) {
+            LOG.info("Attribute type changed for {}.{}: {} -> {}; updating typedef during import",
+                    existing.getName(), incoming.getName(), existingAttribute.getTypeName(), incoming.getTypeName());
```
rkundam: Is incoming.getName() intentional? If incoming and existing have the same name, should this be existing.getName().existingAttribute.getName() instead?

repository/src/test/java/org/apache/atlas/repository/impexp/ImportServiceTest.java
```diff
@@ -364,21 +364,18 @@ public void importGlossary(InputStream inputStream) throws IOException, AtlasBas
...
-            assertNotNull(tag1);
-            assertEquals(tag1.getAllAttributes().size(), 2);
-            throw e;
-        }
+        runImportWithNoParameters(importService, inputStream);
+
+        AtlasClassificationType tag1 = typeRegistry.getClassificationTypeByName("tag1");
+        assertNotNull(tag1);
+        assertEquals(tag1.getAttribute("attrib1").getAttributeDef().getTypeName(), "string");
+        assertNotNull(tag1.getAttribute("attrib2"));
```
rkundam, line 378: Is attrib2 valid attribute name as tag1.json has two attributes attrib1 & attrib3?

saksenasonali: tag1.json is the pre-existing typedef (attrib1/date, attrib3/int). hdfs_path1.zip imports tag1 with attrib1/string and attrib2/int. The test verifies the type change on attrib1 and that attrib2 from the import is added.

rkundam (approved):

PR #671, opened 2026-06-12
ATLAS-5320: Distributed Notification Processing

rkundam:
To support high-volume metadata ingestion while preserving lineage accuracy and data consistency, https://issues.apache.org/jira/browse/ATLAS-5320 introduced a distributed parallel processing architecture. Rather than relying on a single-threaded sequential pipeline, the system partitions entity workloads deterministically and processes independent entity families concurrently.

How was this patch tested?
Tested with different sets of data in clusters and compared with Serial Processing for the same datasets.
Ex: For below dataset, Serial Processing took around 5hrs and Distributed Parallel processing with 3 metadata and 3 lineage topics it took 1.5hrs.
Tables: 6.5K
Column: 130K
Lineage: 1.7K


jackhalfalltrades (approved):

rkundam:
CI/docker-build(push) Build passed but docker setup failed due to known issue with atlas-hbase setup. The change can be merged.

PR #707, opened 2026-07-28
ATLAS-5347: Atlas React UI: Skeleton loader stuck indefinitely in Administration Audit tab when expanding Purge/Auto-Purge entity details

nixonrodrigues:
### **What changes were proposed in this pull request?**

[ATLAS-5347](https://issues.apache.org/jira/browse/ATLAS-5347): Fix Skeleton Loader stuck indefinitely in Administration Audit tab for Purge operations

When expanding the details for PURGE or AUTO_PURGE operations in the React UI Administration Audits tab, the skeleton loader would get stuck spinning indefinitely.

This patch fixes the issue by:

- Explicitly passing loading={false} to <AuditsTab /> within AuditResults.tsx.
- Updating the rendering condition in AttributeProperties.tsx to strictly check for loading === true rather than defaulting to the skeleton loader when loading is undefined.
- Updating the corresponding Jest unit tests in AttributeProperties.test.tsx to match the new strict conditional rendering logic.
- ### How was this patch tested?

- Manual Testing: Verified locally that expanding PURGE and AUTO_PURGE operations in the Administration Audits UI successfully loads the data without freezing the skeleton loader.
- Unit Tests: Updated and successfully ran Jest unit tests (AttributeProperties.test.tsx) to verify that the loader disappears correctly when entity data is available and loading is not explicitly set to true.

### DIFF
--- a/dashboard/src/views/Administrator/Audits/AuditResults.tsx
+++ b/dashboard/src/views/Administrator/Audits/AuditResults.tsx
@@ -232,7 +232,7 @@ const AuditResults = ({ componentProps, row }: any) => {
           button2Handler={undefined}
           maxWidth="lg"
         >
-          <AuditsTab auditResultGuid={currentPurgeResultObj} />
+          <AuditsTab auditResultGuid={currentPurgeResultObj} loading={false} />
         </CustomModal>
       )}
     </>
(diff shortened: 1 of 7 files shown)


pawarprasad123 (approved):
Lchanges look good to me.

PR #717, opened 2026-08-03
ATLAS-5366 - Exclude node_modules from source release tarball

nixonrodrigues:

               Exclude node_modules from source release tarball

## How was this patch tested?

mvn package with -Pdist and verify the atlas source tar ball.


pawarprasad123 (approved):
LGMT

PR #720, opened 2026-08-06
ATLAS-5014 : Basic Search : excludeHeaderAttributes is true, order of…

pareshddevalia:
… attribute in the response changes

Store search attributes in LinkedHashSet so json array order is kept when building attributes.name and attributes.values. Duplicate attribute names are deduped using first-occurrence order. Fixes mismatched column order between request and response for POST /api/atlas/v2/search/basic.

Basic search with excludeHeaderAttributes=true did not preserve the order of attributes from the client request. For example, a request with "attributes": ["qualifiedName", "createTime"] could return "attributes": { "name": ["createTime", "qualifiedName"], ... }, so column order in attributes.values did not match the request.

Root cause: SearchParameters.attributes was stored as Set<String>. Jackson deserialized the JSON array into a HashSet, which does not preserve insertion order. The response builder copied that set into attributes.name and used the same iteration order when building each row in attributes.values.

Fix:

Store attributes in a LinkedHashSet in SearchParameters and QuickSearchParameters
Use @JsonDeserialize(as = LinkedHashSet.class) so JSON deserialization keeps order
Update relationship search REST API to accept List<String> query params, which are normalized through the same setter

## How was this patch tested?

Unit tests
Manual testing
Started Atlas with PostgreSQL + Hive using docker compose
Created hive_table entities via Hive

### DIFF
--- a/intg/src/test/java/org/apache/atlas/model/discovery/TestQuickSearchParameters.java
+++ b/intg/src/test/java/org/apache/atlas/model/discovery/TestQuickSearchParameters.java
@@ -22,8 +22,9 @@ import org.apache.atlas.model.discovery.SearchParameters.FilterCriteria;
 import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.Test;
 
-import java.util.HashSet;
-import java.util.Set;
+import java.util.ArrayList;
+import java.util.LinkedHashSet;
+import java.util.List;
 
 import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertFalse;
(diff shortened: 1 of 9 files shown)


intg/src/test/java/org/apache/atlas/model/discovery/TestQuickSearchParameters.java
```diff
@@ -306,15 +307,15 @@ public void testEmptyAttributesSet() {
...
         attributes.add("initialAttribute");
         searchParameters.setAttributes(attributes);
 
         attributes.add("additionalAttribute");
 
-        assertEquals(searchParameters.getAttributes().size(), 2);
+        assertEquals(searchParameters.getAttributes().size(), 1);
         assertTrue(searchParameters.getAttributes().contains("initialAttribute"));
-        assertTrue(searchParameters.getAttributes().contains("additionalAttribute"));
+        assertFalse(searchParameters.getAttributes().contains("additionalAttribute"));
```
chaitalicod, line 320: Can you check if we need JSON deserialization tests for QuickSearchParameters?
POST /api/atlas/v2/search/quick
similar way we have in TestSearchParameters.java?

pareshddevalia: Added the UT for JSON deserialization tests for QuickSearchParameters.

PR #722, opened 2026-08-10
ATLAS-5370: Atlas UI : Encode text in D3 graph tooltips Lineage and Profile charts

pawarprasad123:

User-provided or API-sourced text shown in D3-based tooltips was inserted into tooltip HTML without encoding. Special characters in entity names, type names, or query text (e.g. <, >, &, quotes) could cause incorrect tooltip rendering or unexpected HTML behavior in the browser.

This PR encodes all dynamic tooltip fields before they are passed to d3-tip().html(), which uses innerHTML under the hood. React auto-escape does not apply here because lineage is rendered via D3, not React JSX.

## Affected areas:

Lineage graph — React UI (dashboard) and classic UI (dashboardv2 atlas-lineage bundle)
Profile bar chart — classic UI only (ProfileBarChart.js)

## Root cause
Both lineage and profile chart components use d3-tip with .html(), which assigns content via innerHTML. Values from the lineage API (displayText → toolTipLabel, typeName, queryText) and profile chart data (d.value, type, d.count) were concatenated directly into HTML strings.

The backend returns raw entity attribute values via EntityGraphRetriever.getDisplayText() without HTML encoding. Classic UI already uses _.escape() in other views (e.g. Utils.js); lineage tooltips did not.

## Why client-side encoding is still required
Even when the API provides displayText, the UI must encode before inserting into HTML tooltips because:

API returns raw entity attribute values — no server-side HTML encoding
d3-tip().html() writes to innerHTML
Classic UI already encodes displayText elsewhere via _.escape()
Encoding only affects tooltip rendering — navigation and graph interactions use GUIDs, not display text

## How was this patch tested?

Automated tests:
1)  npm test -- --testPathPattern="escapeHtml|LineageTab|Lineage"  ->  155/155 passed
2) npm test -- --testPathPattern="escapeHtml" -> 35/35 passed
3) npm run build (dashboard) -> Pass (~12s)
4) mvn apache-rat:check -> Pass (license header on new test file)

Test approach:

Production escapeHtml imported from Utils/index.js (with mocked heavy deps)
Classic UI _.escape behavior verified via parity helper mirroring underscore algorithm
Both paths validated to produce safe tooltip HTML (no raw <script>, , unescaped tags)

Screenshot 2026-08-10 at 4 32 37 PM
Screenshot 2026-08-10 at 4 32 42 PM


farhank31 (approved):
LGTM 👍

chaitalicod:
LGTM +1

chaitalicod (approved):

PR #729, opened 2026-08-18
ATLAS-5375: Fix stale PUT response for /api/atlas/v2/relationship block/unblock propagation

sheetalshah1007:
### Problem

When `atlas.tasks.enabled=true`, relationship PUTs that **change** `propagateTags` or `blockedPropagatedClassifications` queue a background propagation task. Previously, the HTTP 200 body could return **stale** edge state (`blockedPropagatedClassifications`, `propagatedClassifications`) because the response did not reflect the requested block/direction change while entity tag updates ran asynchronously.

Task GUIDs stored on the relationship edge (`__pendingTasks`) had two additional issues:
- They were **not exposed** in relationship JSON, so clients could not see in-flight work.
- They **often did not clear** after COMPLETE/FAILED because **in-place** updates to the edge list property were not persisted (cleanup logic ran, but the graph still held stale GUIDs).

### Solution

This patch updates the relationship **edge synchronously on PUT** and keeps downstream **entity tag propagation** in the background task.

#### Key changes

1. **Synchronous edge updates:** Refactored `DeleteHandlerV1.updateTagPropagations()` to write edge properties (`propagateTags`, blocked IDs) before the API returns. Tasks are queued **only when** `propagateTags` or the blocked list actually changes (no-op PUTs skip task queue).

2. **Pre-change state in task params:** `ClassificationTask` passes `oldTagPropagation` and `oldBlockedClassifications` into the worker so entity tag updates use **pre-PUT** edge state. The worker does not assume the edge is still in its old state (edge is already updated).

3. **API visibility for `pendingTasks`:** Added read-only `pendingTasks` on `AtlasRelationship`. `EntityGraphRetriever` maps edge `__pendingTasks` into JSON when non-empty.

4. **Edge list-property persistence fix:** `AtlasGraphUtilsV2.addItemToListProperty` / `removeItemFromListProperty` now use copy-on-write (`new ArrayList` + `setListProperty`) so queue/remove of task GUIDs persists and `pendingTasks` clears after COMPLETE/FAILED.

### Testing

- Extended `ClassificationPropagationWithTasksTest` and updated related task tests (`ClassificationPropagationTasksTest`, `ClassificationTaskTest`, `ClassificationPropagateTaskFactoryTest`).
- **56** repository tests pass on **Java 8 and Java 17**:

```bash
mvn -pl repository -Dtest=ClassificationPropagationWithTasksTest,ClassificationPropagationTasksTest,ClassificationTaskTest,ClassificationPropagateTaskFactoryTest test
```

Coverage includes:
- Unblock/block PUT and GET show updated edge state **before** task completes
- `pendingTasks` populated while work is in flight; cleared after COMPLETE
- Entity propagation outcomes after background task
- `propagateTags` change wins over blocked change in the same PUT (if/else)

Optional manual check: `dev-support/atlas-scripts/atlas5375_pr_review_test.sh`

### Compatibility & API impact

- **No breaking changes** — additive read-only `pendingTasks` field only.
- **`pendingTasks` omitted when empty** (`@JsonInclude(NON_NULL)`); treat missing field as none in flight.
- **Response shape:** PUT returns relationship at top level; GET wraps in `"relationship"`.
- **Task status:** `GET /api/atlas/admin/tasks?guids={guid}` (admin).
- Legacy in-flight tasks without new params supported via `isLegacyTask`.
- **`propagateTags` required on PUT** — omitting it retains legacy NPE risk (out of scope).

### Trade-off

Edge metadata commits before entity tags. If the background task fails after retries, edge rules may differ from downstream classifications; recover by re-PUT, admin task DELETE + re-PUT, or manual tag correction.

---
## Attachments
**Design doc:** [ATLAS-5375-design-document.md](https://github.com/user-attachments/files/31174862/ATLAS-5375-design-document.md) — implementation notes for reviewers (problem/flow, API behavior, tests, trade-offs). Reference only; not part of the runtime deliverable.

### DIFF
--- a/repository/src/main/java/org/apache/atlas/repository/store/graph/v1/DeleteHandlerV1.java
+++ b/repository/src/main/java/org/apache/atlas/repository/store/graph/v1/DeleteHandlerV1.java
@@ -979,7 +979,7 @@ public abstract class DeleteHandlerV1 {
             return;
         }
 
-        // Step 4: Write propagateTags or blocked list to the edge; skipped on background task unless legacy.
+        // Step 4: Write propagateTags or blocked list to the edge (not both; propagateTags wins if both changed). Skipped on background task unless legacy.
         if (!isAsyncExecution || isLegacyTask) {
             if (propagationChanged) {
                 AtlasGraphUtilsV2.setEncodedProperty(edge, RELATIONSHIPTYPE_TAG_PROPAGATION_KEY, newTagPropagation.name());
@@ -989,7 +989,8 @@ public abstract class DeleteHandlerV1 {
 
             // Step 5: Tasks enabled — edge written above; defer entity tags to background task with pre-change state.
             if (!isAsyncExecution && DEFERRED_ACTION_ENABLED) {
-                createAndQueueTask(CLASSIFICATION_PROPAGATION_RELATIONSHIP_UPDATE, edge, relationship, oldTagPropagation.name(), oldBlocked);
+                createAndQueueTask(CLASSIFICATION_PROPAGATION_RELATIONSHIP_UPDATE, edge, relationship,
+                        oldTagPropagation != null ? oldTagPropagation.name() : null, oldBlocked);
                 return;
             }
         }
--- a/repository/src/main/java/org/apache/atlas/repository/store/graph/v1/DeleteHandlerV1.java
+++ b/repository/src/main/java/org/apache/atlas/repository/store/graph/v1/DeleteHandlerV1.java
@@ -126,6 +126,7 @@ import static org.apache.atlas.repository.store.graph.v2.AtlasGraphUtilsV2.getSt
 import static org.apache.atlas.repository.store.graph.v2.AtlasGraphUtilsV2.isReference;
 import static org.apache.atlas.repository.store.graph.v2.tasks.ClassificationPropagateTaskFactory.CLASSIFICATION_PROPAGATION_ADD;
 import static org.apache.atlas.repository.store.graph.v2.tasks.ClassificationPropagateTaskFactory.CLASSIFICATION_PROPAGATION_DELETE;
+import static org.apache.atlas.repository.store.graph.v2.tasks.ClassificationPropagateTaskFactory.CLASSIFICATION_PROPAGATION_RELATIONSHIP_UPDATE;
 import static org.apache.atlas.type.AtlasStructType.AtlasAttribute.AtlasRelationshipEdgeDirection.OUT;
 import static org.apache.atlas.type.Constants.PENDING_TASKS_PROPERTY_KEY;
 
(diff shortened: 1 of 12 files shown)


repository/src/main/java/org/apache/atlas/repository/store/graph/v1/DeleteHandlerV1.java
```diff
@@ -923,59 +924,106 @@ public void removeFromPropagatedClassificationNames(AtlasVertex entityVertex, St
...
+            entityRetriever.evaluateClassificationPropagation(classificationVertex, propagationsToAdd, propagationsToRemove);
+
+            if (CollectionUtils.isNotEmpty(propagationsToAdd)) {
+                addTagPropagation(classificationVertex, propagationsToAdd);
+            }
+
+            if (CollectionUtils.isNotEmpty(propagationsToRemove)) {
+                removeTagPropagation(classificationVertex, propagationsToRemove);
             }
-        } else {
```
chaitalicod, line 976: kindly check this comment which got removed while removing this block if required bring back the comment

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasRelationshipStoreV2.java
chaitalicod, line 372: handleBlockedClassifications is still present and used only by create now. Please confirm this is intentional (create is synchronous, update is deferred). If so, fine — but the two paths now use two different code paths for the same logical operation

375-379

repository/src/main/java/org/apache/atlas/repository/store/graph/v1/DeleteHandlerV1.java
```diff
@@ -923,59 +924,106 @@ public void removeFromPropagatedClassificationNames(AtlasVertex entityVertex, St
...
+        if (!isAsyncExecution || isLegacyTask) {
+            if (propagationChanged) {
+                AtlasGraphUtilsV2.setEncodedProperty(edge, RELATIONSHIPTYPE_TAG_PROPAGATION_KEY, newTagPropagation.name());
+            } else if (blockedChanged) {
+                setBlockedClassificationIds(edge, classificationIdsToBlock);
+            }
+
+            // Step 5: Tasks enabled — edge written above; defer entity tags to background task with pre-change state.
+            if (!isAsyncExecution && DEFERRED_ACTION_ENABLED) {
+                createAndQueueTask(CLASSIFICATION_PROPAGATION_RELATIONSHIP_UPDATE, edge, relationship, oldTagPropagation.name(), oldBlocked);
```
chaitalicod: NPE guard on oldTagPropagation.name())

chaitalicod:
Can we run this fix on considerable amount of data since we are dealing with deferred actions- Stress testing and performance

repository/src/main/java/org/apache/atlas/repository/store/graph/v1/DeleteHandlerV1.java
sheetalshah1007: Valid - Added a null guard when queuing the task:  getPropagateTags(edge) can return null on legacy edges without the property set; in that case the task is queued without old propagation state and handled as a legacy task.

repository/src/main/java/org/apache/atlas/repository/store/graph/v1/DeleteHandlerV1.java
sheetalshah1007: Restored in the Step 4 comment. Same behavior: we write propagateTags or the blocked list, not both. If both change in one PUT, propagateTags wins.

repository/src/main/java/org/apache/atlas/repository/store/graph/v2/AtlasRelationshipStoreV2.java
sheetalshah1007: Yes, intentional. Create is unchanged — it still runs handleBlockedClassifications and addTagPropagation synchronously. There is no prior edge state on create, and create was not affected by the stale PUT issue. Update writes the edge synchronously and defers entity tag changes via the background task when tasks are enabled. Same propagation logic, different timing for update vs create.

chaitalicod (approved):

PR #752, opened 2026-09-09
ATLAS-5401: client configuration to support custom token supplier

mneethiraj:

Updated Atlas client for applications to register a customer JWT token provider - using configuration `atlas.rest.auth.token.supplier`.

## How was this patch tested?

- added unit tests for the enhancement introduced in this PR
- verified that all existing tests successfully complete


nixonrodrigues (approved):
Changes looks good. I tested the AtlasClient In IntegrationTest with all variants of JWT  token inputs such as  - 
    testBasicAuth
  - testJwtFromFile
  - testJwtFromEnv
  - testCustomTokenSupplier

it  looks good.

```
@Test
    public void testCustomTokenSupplier() throws Exception {
        Configuration configuration = new BaseConfiguration();

        configuration.setProperty("atlas.rest.auth.token.supplier", StaticJwtSupplier.class.getName());

        StaticJwtSupplier.TOKEN = jwtToken;

        AtlasClientV2 client = new AtlasClientV2(configuration, new String[] {ATLAS_URL}, null);

        assertTrue(client.isServerReady(), "isServerReady() should succeed with custom token supplier");
    }

    public static class StaticJwtSupplier implements Supplier<String> {
        static String TOKEN;

        @Override
        public String get() {
            return TOKEN;
        }
    }
```

PR #762, opened 2026-09-17
ATLAS-5409: Atlas Kafka clients authenticate with OAUTHBEARER using K…

nixonrodrigues:
Description
Atlas REST notification and hook consumers fail to publish to Kafka when the cluster uses SASL/OAUTHBEARER with Kubernetes projected service-account tokens (file:///var/run/secrets/kafka/kafka.<audience-id>).

**Symptoms:**

HTTP 500 on POST .../v2/notification/topic/ATLAS_HOOK (often with no useful error body)
Spark–Atlas hook delivery blocked
Logs: scope value must be non-null, token file access errors, or stale token after SA rotation
**Root causes:**

Stock OAuthBearerLoginCallbackHandler rejects K8s SA JWTs (no OAuth scope claim) and may fail File.canRead() on projected token files.
NotificationREST only catches NotificationException; KafkaException surfaces as silent HTTP 500.
Custom/patched handlers that cache the JWT break SA token refresh when kubelet rotates the file in place and Kafka ExpiringCredentialRefreshingLogin re-authenticates.

**Proposed Solution**
1. KubernetesOAuthHandler (new)
Path: notification/src/main/java/org/apache/atlas/security/kafka/KubernetesOAuthHandler.java
Reads JWT via Files.readAllBytes() from file:// URL (sasl.oauthbearer.token.endpoint.url)
No OAuth2 scope validation; broker validates JWT (Strimzi/K8s OIDC in production)
Does not cache tokens — re-reads file on every OAuth handle() (login + refresh)
Re-parses exp / iat / sub each time for correct Kafka refresh scheduling
Retries read/parse during kubelet atomic file replace (5 attempts, 50ms)
Unit tests including token rotation (testHandleReloadsTokenAfterKubernetesRotation)

**2. Error visibility (related)**

NotificationREST: catch (Exception) → AtlasBaseException with real message
KafkaNotification: ERROR log on KafkaProducer creation failure

**4. Configuration (opt-in)**
```
atlas.kafka.security.protocol=SASL_SSL
atlas.kafka.sasl.mechanism=OAUTHBEARER
atlas.kafka.sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required;
atlas.kafka.sasl.oauthbearer.token.endpoint.url=file:///var/run/secrets/kafka/kafka.<audience-id>
atlas.kafka.sasl.login.callback.handler.class=org.apache.atlas.security.kafka.KubernetesOAuthHandler
<audience-id> comes from platform/K8s (projected volume audience / filename), not from Atlas.
```


notification/src/main/java/org/apache/atlas/security/kafka/KubernetesOAuthHandler.java
```diff
@@ -0,0 +1,264 @@
...
+ * {@link #SASL_OAUTHBEARER_TOKEN_ENDPOINT_URL} and supplies it to the
+ * Kafka client without OAuth 2.0 scope validation. The broker validates the JWT.
+ * <p>
+ * Kubernetes rotates projected SA tokens in place (same path, new contents). Kafka's
+ * {@code ExpiringCredentialRefreshingLogin} re-invokes this handler on each login /
+ * refresh; {@link #handle(Callback[])} always re-reads the token file and builds a
+ * new {@link OAuthBearerToken} from the current JWT (including updated {@code exp}).
+ * Token bytes are never cached between callbacks.
+ */
+public class KubernetesOAuthHandler implements AuthenticateCallbackHandler {
```
mneethiraj: I suggest renaming the class from `KubernetesOAuthHandler` to `AtlasKafkaOAuthCallbackHandler`.

nixonrodrigues: Updated the classname to AtlasKafkaOAuthCallbackHandler and tested.

mneethiraj (approved):

